PR #85, opened 2020-06-20
[minor] Readme update

msemelman:
Update readme with better instructions on how to make the docs.

## Description
Replaced current instructions to reflect the latest way to create docs


msemelman:
I did this simple change after getting some errors while following instructions.

vyasr:
@msemelman good catch on realizing there were separate documentation requirements! I'll try and unify these a little more clearly in future releases.

PR #92, opened 2020-10-06
Feature/form factors

vyasr:
## Description

This PR adds the ability to compute the form factor for a shape. The calculation is currently implemented for spheres, polygons, and polyhedra.

## Motivation and Context

The code in ft.py is important to maintain in some form since it represents a relatively unique contribution and may be of future use for purposes of high fidelity diffraction calculations for anisotropic shapes. However, the code quality, style, and correctness could benefit from bringing into the fold of the newer object-base shape API.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/euclid/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/euclid/blob/master/ContributorAgreement.md).
- [x] I have updated the [changelog](https://github.com/glotzerlab/euclid/blob/master/ChangeLog.txt) and the [credits](https://github.com/glotzerlab/euclid/blob/master/doc/source/credits.rst).


vyasr:
@DomFijan I'd be curious to get your thoughts on these calculations. The content of this PR is a means for evaluating the form factor amplitudes for a few shapes (spheres, polygons, and polyhedra). The sphere calculation is very straightforward, but the polyhedron calculation (and by extension, the polygon calculation it uses) were original contributions from one of our old grad students @eirrgang. The code has been shuffled around to various different parts of the glotzerlab code base as we've more clearly delineated the scope and goals for different packages over time. This package focuses on shapes, so it seemed like the best place to retain a reference version of this calculation.

Of course, it's not particularly practical; the implementation here is the amplitude for just a single shape, whereas a realistic form factor calculation would need to accumulate over many particles and then take the complex norm to get the observable intensity. Since that's out of scope for this package, I wanted to maintain this code here as a reference implementation in case we eventually want to implement this directly in something like glotzerlab/freud#660. Having said all that, your perspective on the utility of the code in this form and what might be useful to provide would be helpful.

@bdice you don't really need to review this, I'm just tagging you here so you're aware of it since this is relevant to your ongoing structure factor work and you originally copied this code into coxeter. Review if you'd like, but don't waste your time if you're busy.

coxeter/shape_classes/base_classes.py
```diff
@@ -28,6 +28,38 @@ def gsd_shape_spec(self):
         """dict: Get a :ref:`complete GSD specification <shapes>`."""  # noqa: D401
         return {}
 
+    def compute_form_factor_amplitude(self, q):
+        r"""Calculate the form factor intensity.
+
+        In solid state physics, scattering theory is concerned with understanding the
+        ways in which light is scattered from a sample. For a single point particle at
```
bdice, line 35: Other particles love to scatter too ❤️
```suggestion
        ways in which radiation is scattered from a sample. For a single point particle at
```

coxeter/shape_classes/base_classes.py
```diff
@@ -28,6 +28,38 @@ def gsd_shape_spec(self):
...
+        be identified as the Fourier transform of the total scattering density. When the
+        particles are not point particles, the scattering density of the particles in
+        their local coordinate systems are no longer identical. Conveniently, this
+        component is separable in the Fourier transform of the total density; as a
+        result, the scattering scattering intensity can be decomposed into two terms,
+        the Fourier transform of the distribution of scatterers and the Fourier
+        transform of each scatterer in its local coordinate system. The first term is
+        known as the static structure factor :math:`S(\vec{q})` and describes the
+        spatial distribution of scatterers, while the second term is called the form
+        factor :math:`F(\vec{q})` and describes the local scattering profile.
```
bdice, line 49: Typically lowercase f. Also would love to have Wikipedia links in this docstring.
```suggestion
        factor :math:`f(\vec{q})` and describes the local scattering profile.
```

coxeter/shape_classes/polygon.py
```diff
@@ -480,3 +480,48 @@ def circumcircle(self):
             raise RuntimeError("No circumcircle for this polygon.")
 
         return Circle(np.linalg.norm(x), x + self.vertices[0])
+
+    def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
+        # The form factor amplitude of a polygon is computed according to the
+        # derivation provided in this dissertation:
+        # https://deepblue.lib.umich.edu/handle/2027.42/120906
```
bdice, line 487: Should this be in a docstring / cited in the docs?

coxeter/shape_classes/polygon.py
```diff
@@ -480,3 +480,48 @@ def circumcircle(self):
...
+        ).squeeze()
+        edges_dot_qs = np.inner(edges[:, np.newaxis, :], q_nonzero_broadcast).squeeze()
+        f_ns = (
+            np.dot(edges_cross_qs, self.normal)
+            # Note that np.sinc(x) gives sin(pi*x)/(pi*x)
+            * np.sinc(0.5 * edges_dot_qs / np.pi)
+            / q_sqs[~zero_q]
+        )
+        # Apply translational shift relative to the center of the
+        # polygonal face relative to the polyhedron centroid.
```
bdice, line 522: This seems like an error, but I’m not sure if my suggestion is correct.
```suggestion
        # polygonal face relative to its centroid.
```

coxeter/shape_classes/sphere.py
```diff
@@ -93,3 +93,27 @@ def is_inside(self, points):
         """
         points = np.atleast_2d(points) - self.center
         return np.linalg.norm(points, axis=-1) <= self.radius
+
+    def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
+        # Use the parent docstring.
+
+        # The formula for a the form factor of a sphere may be found here:
+        # http://gisaxs.com/index.php/Form_Factor:Sphere
+        # (among other sources).[jj:w
```
bdice, line 102:
```suggestion
        # (among other sources).
```
Vimtastic

coxeter/shape_classes/sphere.py
```diff
@@ -93,3 +93,27 @@ def is_inside(self, points):
...
+        q = np.atleast_2d(q)
+        form_factor = np.empty(q.shape[0], dtype=np.complex128)
+        qsq = np.sum(q * q, axis=-1)
```
bdice, line 105: Would prefer a consistent name with previous computations of the same thing.
```suggestion
        q_sqs = np.sum(q * q, axis=-1)
```

tests/test_polygon.py
```diff
@@ -313,3 +313,47 @@ def test_incircle_from_center(convex_square):
...
+        1.0,
+        0.95885108,
+        0.95885108,
+        0.84147098,
+        0.95885108,
+        1.0,
+        0.80684536,
+        0.3825737,
+    ]
+    np.testing.assert_almost_equal(square.compute_form_factor_amplitude(ks), ampl)
```
bdice, line 350: Tests should use `assert_allclose` as recommended in the NumPy docs.

bdice (approved):
Some minor comments but looks good overall! Thanks for porting this code over.

coxeter/shape_classes/base_classes.py
vyasr: I thought you were a physicist! From Wikipedia:
> In physics, the term light sometimes refers to electromagnetic radiation of any wavelength, whether visible or not 
but in any case I can change it :)

coxeter/shape_classes/base_classes.py
vyasr: Tbh I've seen a pretty mixed bag, plenty of references use uppercase for all scattering functions e.g. `I(q) = S(q) * F(q)` since they're all Fourier transforms. I'm fine with switching though.

coxeter/shape_classes/polygon.py
vyasr: Yes probably, but at the moment I'm just inheriting the parent docstring, which provides a more detailed description. I'm not sure how best to balance that (I don't want to copy-paste that everywhere), what do you suggest?

coxeter/shape_classes/polygon.py
vyasr: What seems like an error? The interpretation of what the math is doing?

coxeter/shape_classes/sphere.py
vyasr: :)

tests/test_polygon.py
vyasr: Good catch, I just copy-pasted the old tests and changed the computed value without changing that.

coxeter/shape_classes/polygon.py
bdice: The arguments are different for the child class, so I think this would need a proper docstring of its own, regardless. I would just put in a reference link to the parent method for basic info and specify additional information (like the algorithm source and additional arguments) in this child class docstring.

coxeter/shape_classes/polygon.py
vyasr: Discussed offline, I didn't realize this was in the polygon class :) this correction makes sense.

bdice:
@vyasr FYI looks like there was an unexpected test failure here: https://app.circleci.com/pipelines/github/glotzerlab/coxeter/291/workflows/a160cbff-7514-41ff-b36b-6d2e9b9ceced/jobs/929

DomFijan (approved):

coxeter/shape_classes/base_classes.py
```diff
@@ -28,6 +28,42 @@ def gsd_shape_spec(self):
...
+        point particles, the scattering density of the particles in their local
+        coordinate systems are no longer identical. Conveniently, this component is
+        separable in the Fourier transform of the total density; as a result, the
+        scattering scattering intensity can be decomposed into two terms, the Fourier
+        transform of the distribution of scatterers and the Fourier transform of each
+        scatterer in its local coordinate system. The first term is known as the
+        `static structure factor <https://en.wikipedia.org/wiki/Structure_factor>`__
+        :math:`S(\vec{q})` and describes the spatial distribution of scatterers, while
+        the second term is called the
+        `form factor <https://en.wikipedia.org/wiki/Atomic_form_factor>`__
```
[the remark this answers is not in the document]
vyasr: Yes that's a good point. I prefer GISAXS, so I linked to http://gisaxs.com/index.php/Form_Factor for now. The other two links you proposed are strictly for polytopes, so I don't think this is quite the right place for it.

coxeter/shape_classes/polygon.py
vyasr: This is done.

PR #99, opened 2020-10-08
Refactor/modify families api

vyasr:
## Description

Modify the shape families API to work entirely with classes (not instances) using the `get_shape` `classmethod` rather than a call operator.

## Motivation and Context

This change simplifies the API since it prevents users from having to instantiate a family (`PlatonicFamily()('Cube')`), which could be confusing to users. It also simplifies documentation since we don't have to document a call operator. There are a couple of minor downsides with the new approach. First, to support easily creating new tabulated shape classes I added factory methods, but because the classes are not meant for instantiation these are now _class_ factories (i.e. they generate new subclasses rather than instances). This implementation detail is transparent to users of the API, but may be slightly confusing to developers (although I've kept it as absolutely minimal as possible). Second, since properties are defined on instances, not classes, the raw data for the tabulated shape classes is now directly exposed rather than through a property. I don't view this as a significant change; since the data was a dictionary, nothing really prevented it from being mutable in the old scenario anyway. While there are ways around this, I think the level of sophistication required to achieve them introduce an unjustifiable maintenance burden, so I think the current approach is the best tradeoff.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


coxeter/shape_families/plane_shape_families.py
```diff
@@ -322,7 +269,151 @@ class Family523(TruncationPlaneShapeFamily):
     S = golden_ratio
     """The constant S (the golden ratio)."""
 
-    def __call__(self, a, c):
+    planes = np.array(
```
b-butler, line 272: The documentation of `planes` and `plane_types` gets ugly. Also, there is nothing preventing an user for messing this up. I don't want to go so far as to implement class level properties through metaclasses, but it may be worth being unpythonic and using get methods for these attributes so you can return copies.

This would also prevent the full value of the attribute being shown in the documentation which seems a bit much.

b-butler (changes requested):
Overall I like this API more, and I don't think it is overly complicated the returning of subclasses in the tabulated shape family is the most difficult thing to understand and it isn't that bad. I did leave one comment regarding the plane shape family.

b-butler:
Also, I noticed that the docs raise a lot of warnings when building.

coxeter/shape_families/plane_shape_families.py
vyasr: I agree, I think this is a good call.

PR #100, opened 2020-10-08
Assorted minor fixes

bdice:
## Description

Lots of fixes to documentation, arguments, return types, LaTeX syntax, and spelling.

## Types of Changes

- [x] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:


coxeter/shape_classes/convex_polygon.py
```diff
@@ -116,7 +116,7 @@ def __init__(self, vertices, normal=None, planar_tolerance=1e-5):
 
     @property
     def incircle_from_center(self):
-        """`coxeter.shape_classes.Circle`: Get the largest concentric inscribed circle.
+        """:class:`~.Circle`: Get the largest concentric inscribed circle.
```
Tobias-Dwyer, line 119: I think I've just never seen this syntax before, could you point me to where I can read more about what's going on here :class:`~.Circle`

bdice: I changed this to match the shorthand convention used elsewhere in the code. Basically the `~` just makes it appear as "Circle" instead of "coxeter.shape_classes.Circle". The namespace is assumed to match the namespace of the file where this is defined, I believe. I can't find a good reference on how `~` works, since that's hard to search online...

Tobias-Dwyer (commented):
There was just one syntax change I didn't really understand, which I commented on below. Otherwise, looks good. (maybe you can fill me in on how where to read about that?? ) :D

coxeter/shape_classes/convex_polygon.py
Tobias-Dwyer: hey, thanks! That makes sense

vyasr, line 119: This change actually isn't necessary because of how I've configured Sphinx (see [the rendered docs](https://coxeter.readthedocs.io/en/latest/package-shape-classes.html#coxeter.shape_classes.ConvexPolygon.incircle_from_center)). Technically we could go through and change everything to remove the roles everywhere, but I don't think that's a worthwhile use of time. We can let this change through so you don't have to undo it, just an FYI for future work.

vyasr (changes requested):
Thanks! A few minor changes, then this should be good to go.

Tobias-Dwyer (approved):
These changes look good.

coxeter/shape_classes/convex_polygon.py
vyasr: @Tobias-Dwyer @bdice it's documented [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-syntax).

coxeter/shape_families/common_families.py
```diff
@@ -29,7 +29,7 @@ def __call__(self, n):
 
         Args:
             n (int):
-                An integer (greater than 3)s
+                An integer greater than or equal to 3.
```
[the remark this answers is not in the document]
bdice: Reverted!

coxeter/shape_families/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 Shape families are coxeter's way of providing well-defined methods for
 generating classes of shapes according to some set of rules. The basic
-interface is defined by the :class:`~.ShapeFamily`, which is a functor that is
+interface is defined by the :class:`~._ShapeFamily`, which is a functor that is
```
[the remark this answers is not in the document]
bdice: Reverted!

coxeter/shape_classes/convex_polygon.py
bdice: It made a difference in how things were rendered. I wasn't able to make it link without this. Also this matches the other docstrings returning class types. 🤷‍♂️

bdice: These choices of syntax work.

Rendered as full name "coxeter.shape_classes.Circle":
```
:class:`coxeter.shape_classes.Circle`
`coxeter.shape_classes.Circle`
```

Rendered as short name "Circle":
```
:class:`~.Circle`
`Circle`
```

@vyasr and I decided offline that we prefer the last one because it is shortest. In a future PR we will drop all `:class:` directives and use short names for class references.

vyasr (approved):

Issue #91, opened 2020-10-01
Implement IQ in Polygon class


syjlee:
### Feature description
To implement IQ in shape_classes.Polygon, which is currently raising notImplemented error.

### Proposed solution
area function is already included. Will add perimeter function and iq.

PR #107, opened 2020-10-09
Feature/more abstract methods

### LINKED ISSUE #98: Shape2D Abstract Properties (perimeter, planar moments of inertia)
### Feature description

I noticed that there are some abstract properties of `Shape3D` that are not mirrored for `Shape2D`.

`Shape3D` defines abstract volume, surface area, and inertia tensor properties.

`Shape2D` defines only area.

### Proposed solution

I suggest adding perimeter and planar moments of inertia to the `Shape2D` abstract properties.

I believe the perimeter property is not yet implemented for polygon or its convex/sphero variants.

vyasr:
## Description

Expands the minimum expected API for anything to be a shape. In the process defined perimeter calculations for polygons and spheropolygons, and got a few other properties for free across the board by moving general methods up the hierarchy (for instance iq).

## Motivation and Context

Resolves #98 and #91. I exercised some judgment in which properties _have_ to be implemented (i.e. which are abstract) and which ones _should_ be implemented (concrete methods/properties that raise `NotImplementedError`). I tried to be pragmatic with this decision more than anything else. I also removed the abstractmethod setters since they don't do anything; we'll rely on the tests @Tobias-Dwyer is writing to manually verify these.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -10,6 +10,7 @@ Added
 
 -  Circumsphere and insphere from center calculations for ConvexSpheropolyhedron.
 -  Shape families associated with a DOI can be directly accessed via a dictionary.
+-  Expected abstract interface for shapes (both 2D and 3D) has expanded.
 
 Changed
 ~~~~~~~
--- a/coxeter/shape_classes/base_classes.py
+++ b/coxeter/shape_classes/base_classes.py
@@ -66,10 +66,10 @@ class Shape2D(Shape):
         axes. In addition to the two planar moments, this property also
         provides the product of inertia.
 
-        The `planar moments <https://en.wikipedia.org/wiki/Polar_moment_of_inertia>`__
+        The `planar moments of inertia <https://en.wikipedia.org/wiki/Polar_moment_of_inertia>`__
         and the
-        `product <https://en.wikipedia.org/wiki/Second_moment_of_area#Product_moment_of_area>`__
-        of inertia define the in-plane area distribution.
+        `product of inertia <https://en.wikipedia.org/wiki/Second_moment_of_area#Product_moment_of_area>`__
+        define the in-plane area distribution.
         """  # noqa: E501
         raise NotImplementedError
 
@@ -87,8 +87,21 @@ class Shape2D(Shape):
 
     @property
     def iq(self):
-        """float: The isoperimetric quotient."""
-        # TODO: Add reference to math and some resource in the docstring.
+        r"""float: The isoperimetric quotient.
+
+        The `isoperimetric quotient
+        <https://mathworld.wolfram.com/IsoperimetricQuotient.html>`__ is the ratio of
+        the area of a shape to the area of a circle with the same perimeter. Given a
+        shape of area :math:`A` and perimeter :math:`p`, the circle with the same
+        perimeter has radius :math:`r_p = \frac{p}{2\pi}` and therefore has an area
+        :math:`A_{circle} = \pi r_p^2 = \frac{p^2}{4\pi}`. Therefore, we have that:
+
+        .. math::
+            \begin{align}
+                IQ &= \frac{A}{A_{circle}} \\
+                   &= \frac{4\pi A}{p^2}
+            \end{align}
+        """  # noqa: E501
         return 4 * np.pi * self.area / (self.perimeter ** 2)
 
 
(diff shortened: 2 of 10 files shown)


ChangeLog.rst
```diff
@@ -10,6 +10,7 @@ Added
 
 -  Circumsphere and insphere from center calculations for ConvexSpheropolyhedron.
 -  Shape families associated with a DOI can be directly accessed via a dictionary.
+-  Expected abstract interface for shapes (both 2D and 3D) has expanded.
```
bdice, line 13: This changelog format is a little weird, we should not need two spaces after the dash. I notice there's inconsistency with the "Fixed" section below.

coxeter/shape_classes/base_classes.py
```diff
@@ -18,15 +20,27 @@ def center(self):
...
-    @center.setter
-    @abstractmethod
-    def center(self, value):
-        pass
-
     @property
     def gsd_shape_spec(self):
         """dict: Get a :ref:`complete GSD specification <shapes>`."""  # noqa: D401
-        return {}
+        raise NotImplementedError
```
bdice, line 26: Good, I was going to ask about this.

coxeter/shape_classes/base_classes.py
```diff
@@ -38,11 +52,45 @@ def area(self):
...
+        r"""list[float, float, float]: Get the planar and product moments of inertia.
+
+        Moments are computed with respect to the :math:`x` and :math:`y`
+        axes. In addition to the two planar moments, this property also
+        provides the product of inertia.
+
+        The `planar moments <https://en.wikipedia.org/wiki/Polar_moment_of_inertia>`__
+        and the
+        `product <https://en.wikipedia.org/wiki/Second_moment_of_area#Product_moment_of_area>`__
+        of inertia define the in-plane area distribution.
```
bdice, line 72: I would prefer to give the full term in the link title: `planar moments of inertia <...>` and `product moment of area <...>`.

coxeter/shape_classes/base_classes.py
```diff
@@ -38,11 +52,45 @@ def area(self):
...
+        normal vector).
+
+        The polar moment is computed as the sum of the two planar moments of inertia.
+        """  # noqa: E501
+        return np.sum(self.planar_moments_inertia[:2])
+
+    @property
+    def iq(self):
+        """float: The isoperimetric quotient."""
+        # TODO: Add reference to math and some resource in the docstring.
```
bdice, line 91: I recommend we solve this TODO now. Here's two links I like:
1. https://mathworld.wolfram.com/IsoperimetricQuotient.html
2. https://en.wikipedia.org/wiki/Isoperimetric_inequality#On_a_plane

coxeter/shape_classes/base_classes.py
```diff
@@ -53,23 +101,18 @@ def volume(self):
...
-    @abstractmethod
     def inertia_tensor(self):
         """:math:`(3, 3)` :class:`numpy.ndarray`: Get the inertia tensor."""
-        pass
+        raise NotImplementedError
+
+    @property
+    def iq(self):
+        """float: The isoperimetric quotient."""
+        # TODO: allow for non-spherical reference ratio (changes the prefactor)
```
bdice, line 117: Isn't using a non-spherical reference the same as taking the ratio of two IQs that both use the spherical reference? That is, isn't this a trivial thing to ask users to factor out?

(Reminds me of the change of base formula, `log10(x) = ln(x)/ln(10)`, for some reason.)

tests/test_polygon.py
```diff
@@ -313,3 +313,22 @@ def test_incircle_from_center(convex_square):
     circle = convex_square.incircle_from_center
     assert np.all(circle.center == convex_square.center)
     assert circle.radius == 0.5
+
+
+@pytest.mark.parametrize("num_sides", range(3, 6))
+def test_perimeter(num_sides):
+    """Test the polygon perimeter calculation."""
+
+    def unit_area_regular_ngon_sidelength(n):
```
bdice, line 322:
```suggestion
    def unit_area_regular_n_gon_side_length(n):
```

tests/test_polygon.py
```diff
@@ -313,3 +313,22 @@ def test_incircle_from_center(convex_square):
...
+        r"""Compute the side length of a unit-area regular polygon analytically.
+
+        The area of regular n-gon is given by
+        :math:`\\frac{s^2 n}{4 \\tan\\left(\\frac{180}{n}\\right)}`. This function sets
```
bdice, line 326: raw string, too many slashies ⚔️
```suggestion
        :math:`\frac{s^2 n}{4 \tan\left(\frac{180}{n}\right)}`. This function sets
```

bdice (changes requested):

ChangeLog.rst
vyasr: I agree that it's strange, tbh I assumed you had changed it to this formatting at some point and didn't feel like doing anything about it. We can revise it later.

coxeter/shape_classes/base_classes.py
vyasr: Yes, it is. I think someone (maybe @tcmoore3) made this request, which is how the TODO got added.

ChangeLog.rst
vyasr: #108

bdice (approved):
LGTM!

PR #109, opened 2020-10-09
Enable doctests

### LINKED ISSUE #104: Example docstrings aren't rendered properly
### Description

The syntax for embedded examples isn't quite right. It needs a few changes to match standards for [Google style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html):
1. Need a newline above the word "Example:" (one colon, not two).
2. Need to wrap input lines as if typing into an interpreter with line continuation `...`:
```python
>>> print('This is a long expression and it is split across',
... 'two lines. Note the periods at the beginning of this line.')
```
3. Some examples are a little lengthy and could be trimmed (do we need to get an incircle of a polygon _and_ compute its area?)
4. Write example code in a way that would pass through a formatter (e.g. black). For example, put spaces after commas in all lists of items:
```python
vertices = [[1, 1], [1, -1], [-1, -1], [-1, 1]]
# instead of
vertices=[[1,1],[1,-1],[-1,-1],[-1,1]]
```

 Here's what it looks like:

Incorrect syntax:

```python
class ConvexPolygon(Polygon):
    """A convex polygon.

    The polygon is embedded in 3-dimensions, so the normal
    vector determines which way is "up".

    Args:
        ...
    Example::
        >>> square = coxeter.shape_classes.ConvexPolygon([[1,1],[-1,-1],
                                                          [1,-1],[-1,1]])
        >>> square.area
        4.0
        >>> # ...continued
    """
```

Fixed syntax:

```python
class ConvexPolygon(Polygon):
    """A convex polygon.

    The polygon is embedded in 3-dimensions, so the normal
    vector determines which way is "up".

    Args:
        ...

    Example:
        >>> square = coxeter.shape_classes.ConvexPolygon(
        ...   [[1, 1], [-1, -1], [1, -1], [-1, 1]])
        >>> square.area
        4.0
        >>> # ...continued
    """
```

Incorrect output:
image

Fixed output:
image

bdice:
## Description
This PR enables doctests in pytest, so they will run during continuous integration.

To make doctests work, I enabled two options.

`ELLIPSIS` allows non-deterministic results to be ignored by putting `...`.
`NORMALIZE_WHITESPACE` allows lines to be wrapped for flake8 even if the output does not contain newlines.

I also reformatted the examples sections so they render better. Resolves #104.

## Motivation and Context
Doctests ensure that the code examples continue to work as the package is changed.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
             python -m flake8 --show-source coxeter/
             # Ignore isort errors in tests that are due to ambiguity over whether or not coxeter is third party.
             python -m flake8 --extend-ignore=I001,I004 --show-source tests/
-            python -m pytest --doctest-modules
+            python -m pytest
             python --version
             python -m pip --version
             python -m pip install --progress-bar off --user -U twine wheel setuptools
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
             python -m pip install .[test] --progress-bar off --user
             python -c "import numpy; print('numpy', numpy.__version__)"
             python -c "import scipy; print('scipy', scipy.__version__)"
-            python -m pytest --doctest-modules
+            python -m pytest
 
   test-3.7:
     <<: *test-template
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
             python -m pip install .[test] --progress-bar off --user
             python -c "import numpy; print('numpy', numpy.__version__)"
             python -c "import scipy; print('scipy', scipy.__version__)"
-            pytest
+            python -m pytest --doctest-modules
 
   test-3.7:
     <<: *test-template
@@ -66,7 +66,7 @@ jobs:
             python -m flake8 --show-source coxeter/
             # Ignore isort errors in tests that are due to ambiguity over whether or not coxeter is third party.
             python -m flake8 --extend-ignore=I001,I004 --show-source tests/
-            python -m pytest
+            python -m pytest --doctest-modules
             python --version
             python -m pip --version
             python -m pip install --progress-bar off --user -U twine wheel setuptools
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
             python -m flake8 --show-source coxeter/
             # Ignore isort errors in tests that are due to ambiguity over whether or not coxeter is third party.
             python -m flake8 --extend-ignore=I001,I004 --show-source tests/
-            python -m pytest
+            python -m pytest --doctest-modules
```
vyasr, line 69:
```suggestion
            python -m pytest
```
This change is redundant now.

setup.cfg
```diff
@@ -39,4 +39,5 @@ known_third_party = hypothesis,pytest
 known_first_party = conftest,coxeter,utils
 
 [tool:pytest]
-addopts = -p coxeter.__doctest_fixtures
+addopts = --doctest-modules -p coxeter.__doctest_fixtures
```
vyasr, line 42: This doesn't have to be done in this PR, but in hindsight I'm wondering whether doing everything in terms of the top-level coxeter namespace is really preferable. Should we just use the local names everywhere (e.g. `circle = Circle(...)` instead of `circle = coxeter...`)? If so, then we could get rid of the extra fixtures entirely.

vyasr (changes requested):
While you're doing this, can you also delete all the extra examples as mentioned in #104? For instance, it's enough to do `circumcircle = square.circumcircle` without asking for `circumcircle.radius` etc in the example. Other than that it looks good!

setup.cfg
bdice: I like having full names in the code examples, because they're easier to copy-paste. I support leaving it as-is.

vyasr (approved):
Thanks!

PR #111, opened 2020-10-09
added getter and setters for a and b

Tobias-Dwyer:
Adding getters and setters for the a and b parameters of an ellipsoid. 
## Description

See summary. 
## Motivation and Context

There were no getter or setter checks for a and b. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/coxeter/shape_classes/ellipse.py
+++ b/coxeter/shape_classes/ellipse.py
@@ -47,6 +47,10 @@ class Ellipse(Shape2D):
     """
 
     def __init__(self, a, b, center=(0, 0, 0)):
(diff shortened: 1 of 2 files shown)


bdice (changes requested):
Does this PR need to include validation on the setters? (If so, please also test that errors are raised by invalid values.)
```python
    @a.setter
    def a(self, a):
        if a > 0:
            self._a = a
        else:
            raise ValueError("The semi-major axis a must be positive.")
```

vyasr (approved):

PR #113, opened 2020-10-09
added radius getter setter tests

Tobias-Dwyer:
added radius getter setter tests and check that the radius is positive. 
## Description

added radius getter setter tests and check that the radius is positive. 
## Motivation and Context

These getter and setter tests do were not in the code before. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/tests/test_spheropolygon.py
+++ b/tests/test_spheropolygon.py
@@ -36,6 +36,16 @@ def test_2d_verts(square_points):
     ConvexSpheropolygon(square_points, 1)
 
 
+@given(floats(0.1, 1000))
+def radius_getter_setter_tests(r, square_points):
+    """Test getting and setting the radius."""
+    square_points = square_points[:, :2]
+    convexspheropolygon = ConvexSpheropolygon(square_points, r)
+    assert convexspheropolygon.radius == r
+    convexspheropolygon.radius = r + 1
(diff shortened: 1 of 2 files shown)


tests/test_spheropolygon.py
```diff
@@ -36,6 +36,16 @@ def test_2d_verts(square_points):
...
 
 
+@given(floats(0.1, 1000))
+def radius_getter_setter_tests(r, square_points):
+    """Test getting and setting the radius."""
+    square_points = square_points[:, :2]
+    convexspheropolygon = ConvexSpheropolygon(square_points, r)
+    assert convexspheropolygon.radius == r
+    convexspheropolygon.radius = r + 1
+    assert convexspheropolygon.radius == r + 1
```
bdice, line 46: Testing your error behavior is also a good idea.
```suggestion
    assert convexspheropolygon.radius == r + 1
    with pytest.raises(ValueError):
        convexspheropolygon.radius = 'invalid value'
```

vyasr:
@Tobias-Dwyer it would be good to change the constructor to take advantage of the validation in the radius property. Instead of checking the radius value in the constructor and then setting `self._radius`, just set `self.radius` so that the property setter checks for you.

vyasr (approved):

PR #123, opened 2020-10-14
Make a Python 3 wheel, not a universal wheel.

bdice:
## Description
Coxeter isn't compatible with Python 2. Thus, we can't create a "universal" wheel. Instead, we need to use a Python 3 tag.
https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels

@vyasr You could consider deleting the previous wheel for 0.4.0 and re-generating a wheel with this fix. (Or you can decide that it doesn't matter.)

## Motivation and Context
Fixes wheel compatibility.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ tag = True
 message = Bump up to version {new_version}.
 
 [bdist_wheel]
-universal = 1
+python-tag = py3
 
 [bumpversion:file:coxeter/__init__.py]
 


vyasr:
I'm going to decide that it doesn't matter. If someone actually tries to install on Python2 and complains we can deal with it. But thanks for the fix! I hadn't thought about the fact that I've actually made changes that make this Python2-incompatible.

vyasr (approved):

PR #124, opened 2020-10-17
More consistent setters

### LINKED ISSUE #114: Make setter names consistent
The setters for Python properties should consistently either use the name of the property or just `value` as the argument. This will improve code consistency and readability.

bdice:
## Description
I made property setters look the same:
- Input parameter is named `value`.
- Property setting logic is always first (`if value > 0:`), error raising is always second (`else: raise`). This is my preference because it ensures the expected path is first (easier to read) and the condition matches the error message text (`if value > 0` matches the error `Value must be greater than zero.` and is thus easier to think through).

I also added setters for ellipse area and ellipsoid volume. (I specifically needed ellipsoid volume but wanted ellipse area to match.)

## Motivation and Context
Resolves #114. I did not know the issue already existed and had been assigned to @Tobias-Dwyer, I just wanted to add the volume setter for ellipsoids and figured I'd improve the code base at the same time.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,6 +1,13 @@
 The format is based on `Keep a Changelog <http://keepachangelog.com/en/1.0.0/>`__.
 This project adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`__.
 
+next
+----
+
+Added
+~~~~~
+
+- Ellipse area setter and Ellipsoid volume setter.
 
 v0.4.0 - 2020-10-14
 -------------------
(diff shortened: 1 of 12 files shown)


vyasr:
@Tobias-Dwyer this seems like a good PR for you to review :)

Tobias-Dwyer (approved):
All changes look Rolex to me :D . Thanks for adding the volume setters. Can you do the same thing for setting the surface area? just scale it? (if so I could probably go in and add it)

bdice:
@Tobias-Dwyer Let’s open an issue and do surface areas / perimeters / etc. in a separate PR? I think the scope of this is complete as it is.

Tobias-Dwyer:
Yep, I agree :D

bdice:
@Tobias-Dwyer Issue created: https://github.com/glotzerlab/coxeter/issues/125

PR #131, opened 2020-12-18
Add _rescale for use with setters.

bdice:
## Description
Use `_rescale` to change the length scale of all shapes. The method is implemented for each shape class, and allows for common logic for setters for properties like areas and volumes.

I also added several new property setters.

## Motivation and Context
Simplifies implementations of property setters.

@Tobias-Dwyer Would you like to add tests to this PR? Tests are missing for the new setters. You can model them after similar property tests for other classes. After you've tested them, we can mark them off on the Feature Matrix document.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/coxeter/shapes/base_classes.py
+++ b/coxeter/shapes/base_classes.py
@@ -20,6 +20,16 @@ class Shape(ABC):
         """:math:`(3, )` :class:`numpy.ndarray` of float: Get or set the centroid of the shape."""  # noqa: E501
         pass
 
+    @abstractmethod
+    def _rescale(self, scale):
+        """Multiply length scale.
+
+        Args:
+            scale (float):
+                Scale factor.
+        """
+        pass
+
     @property
     def gsd_shape_spec(self):
         """dict: Get a :ref:`complete GSD specification <shapes>`."""  # noqa: D401
--- a/coxeter/shapes/circle.py
+++ b/coxeter/shapes/circle.py
@@ -68,6 +68,15 @@ class Circle(Shape2D):
         else:
             raise ValueError("Radius must be greater than zero.")
 
+    def _rescale(self, scale):
+        """Multiply length scale.
+
+        Args:
+            scale (float):
+                Scale factor.
+        """
+        self.radius *= scale
+
     @property
     def area(self):
         """float: Get the area of the circle."""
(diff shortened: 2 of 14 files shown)


vyasr:
The `rescale` method also needs a slightly more descriptive docstring explaining that the "length scale" for different shapes means different things, and that the rescaling is a strict linear rescaling.

bdice:
> The `rescale` method also needs a slightly more descriptive docstring explaining that the "length scale" for different shapes means different things, and that the rescaling is a strict linear rescaling.

@vyasr Can you elaborate on the kind of comment you’re expecting? I don’t necessarily agree with your statement that the length scale means different things for different shapes. Length is a 1D quantity. Would a rename to `_rescale_length` help clarify this?

vyasr:
> > The `rescale` method also needs a slightly more descriptive docstring explaining that the "length scale" for different shapes means different things, and that the rescaling is a strict linear rescaling.
> 
> @vyasr Can you elaborate on the kind of comment you’re expecting? I don’t necessarily agree with your statement that the length scale means different things for different shapes. Length is a 1D quantity. Would a rename to `_rescale_length` help clarify this?

Sure, basically what I mean is that `_rescale` is an internal method, not part of the API, and since it's abstract subclasses need to implement it. Therefore, the docstring should provide enough information to make clear what that entails for developers. There is no `self._length_scale` property for shapes, implementers need to set the appropriate property for a particular class of shape and callers need to understand that that's what the method does. I would be happy with an extra sentence containing concrete examples, i.e. "For example, this rescaling should be applied to the radius of a sphere, or to the vertices of a polygon." Otherwise it won't be clear to future developers why they should use this method, and they'll just manually write `self.vertices *= scale` if they add anything that requires a rescaling (and honestly that's probably going to happen anyway for simple shapes like discs).

bdice (commented):
@glotzerlab/coxeter-maintainers I reviewed the features introduced in this PR. I believe all the new setters are covered by the new tests (thanks @Tobias-Dwyer). This should be ready to review and merge (I can't approve my own PR).

Tobias-Dwyer (approved):
Looks good!

PR #132, opened 2020-12-19
Add to_plato_scene feature.

bdice:
[plato_scenes.pdf](https://github.com/glotzerlab/coxeter/files/5927311/plato_scenes.pdf)

## Description
Allows shapes to be displayed as [plato](https://plato-draw.readthedocs.io/en/latest/) scenes.

Most shapes are shown at `self.center`. Polygon-like shapes must be shown in a 2D scene, and thus must be rotated so that their normal is `[0, 0, 1]` and their center is the origin. Orientations are assumed to be `[1, 0, 0, 0]`. Colors are set to 50% gray.

## Motivation and Context
Possible replacement for the `plot` method.

See the attached file for rendered examples!

@glotzerlab/coxeter-maintainers how do we want to test this? Add optional tests that function if plato is installed? If so, there is another issue that backends might still not be installed, so this is just hard to test on CI. In garnett, we have [a pretty hacky workaround](https://github.com/glotzerlab/garnett/blob/f9cb7bad391299e28feb4010eb77447fdc4512cb/tests/test_plato_scenes.py#L79-L92).

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/coxeter/families/tabulated_shape_family.py
+++ b/coxeter/families/tabulated_shape_family.py
@@ -4,8 +4,8 @@
 """Define tabulated shape families.
 
 Tabulated shape families are defined by a dictionary of data that can be
-converted into a shape by some well-defined schema. One example is the
-`GSD shape schema <shapes>`, which defines how to translates dictionaries with
+converted into a shape by some well-defined schema. One example is the `GSD
+shape schema <gsd:shapes>`, which defines how to translates dictionaries with
 the appropriate formatting into a shape. These shape families may be
 constructed from a JSON file that can be read into a dictionary with the
 appropriate formatting.
@@ -90,7 +90,7 @@ class TabulatedGSDShapeFamily(TabulatedShapeFamily):
     """A tabulated shape family defined by a GSD shape schema.
 
     The values of the dictionary used to construct this class must adhere to
-    the :ref:`GSD shape spec <shapes>`. Each mapping may contain additional
+    the :ref:`GSD shape spec <gsd:shapes>`. Each mapping may contain additional
     data, which is ignored when the class is called to actually produce
     :class:`~coxeter.shapes.Shape` objects.
 
--- a/coxeter/shape_getters.py
+++ b/coxeter/shape_getters.py
@@ -27,9 +27,9 @@ def from_gsd_type_shapes(params, dimensions=3):  # noqa: C901
     See :ref:`here <gsd:shapes>` for the specification of the schema. Note that
     the schema does not differentiate between 2D and 3D shapes for spheres (vs.
     circles) and ellipsoids (vs. ellipses) because in context the
-    dimensionality of the shape can be inferred from simulation boxes. To
-    address this ambiguity, this function accepts a dimensions parameter that
-    can be used to disambiguate explicitly between these two cases.
+    dimensionality of those shapes can be inferred from simulation boxes. To
+    address this ambiguity, this function accepts a ``dimensions`` parameter
+    that can be used to disambiguate explicitly between these two cases.
 
     Args:
         params (dict):
--- a/coxeter/shape_getters.py
+++ b/coxeter/shape_getters.py
@@ -24,12 +24,12 @@ from .shapes import (
 def from_gsd_type_shapes(params, dimensions=3):  # noqa: C901
     """Create a :class:`~.Shape` from a dict conforming to the GSD schema.
 
-    See :ref:`here <gsd:shapes>` for the specification of the schema. Note
-    that the schema does not differentiate between 2D and 3D shapes for
-    spheres (vs. circles) and ellipsoids (vs. ellipses) because in context
-    those can be inferred from simulation boxes. To address this ambiguity,
-    this function accepts a dimensions parameter that can be used to
-    disambiguate explicitly between these two cases.
+    See :ref:`here <gsd:shapes>` for the specification of the schema. Note that
+    the schema does not differentiate between 2D and 3D shapes for spheres (vs.
+    circles) and ellipsoids (vs. ellipses) because in context the
+    dimensionality of the shape can be inferred from simulation boxes. To
+    address this ambiguity, this function accepts a dimensions parameter that
+    can be used to disambiguate explicitly between these two cases.
 
     Args:
         params (dict):
--- a/coxeter/shape_getters.py
+++ b/coxeter/shape_getters.py
@@ -24,12 +24,12 @@ from .shapes import (
 def from_gsd_type_shapes(params, dimensions=3):  # noqa: C901
     """Create a :class:`~.Shape` from a dict conforming to the GSD schema.
 
-    See :ref:`here <shapes>` for the specification of the schema. Note that the
-    schema does not differentiate between 2D and 3D shapes for spheres (vs
-    circles) and ellipsoids (vs ellipses) because in context those can be
-    inferred from simulation boxes.  To address this ambiguity, this function
-    accepts a dimensions parameter that can be used to disambiguate explicitly
-    between these two cases.
+    See :ref:`here <gsd:shapes>` for the specification of the schema. Note
+    that the schema does not differentiate between 2D and 3D shapes for
+    spheres (vs. circles) and ellipsoids (vs. ellipses) because in context
+    those can be inferred from simulation boxes. To address this ambiguity,
+    this function accepts a dimensions parameter that can be used to
+    disambiguate explicitly between these two cases.
 
     Args:
         params (dict):
--- a/coxeter/shapes/base_classes.py
+++ b/coxeter/shapes/base_classes.py
@@ -133,6 +133,41 @@ class Shape(ABC):
     def __str__(self):
         return repr(self)
 
+    def _plato_primitive(self, backend, **kwargs):
+        raise NotImplementedError("This shape does not have a plato primitive.")
+
+    def to_plato_scene(self, backend=None, scene=None):
+        """Create a plato scene displaying this shape.
+
+        Args:
+            backend (str):
+                Name of backend to use from plato. The backend must support
+                the primitive corresponding to this shape. Defaults to None,
+                in which case the ``scene`` argument must be provided.
+            scene (:class:`plato.draw.Scene`):
+                Scene object to render into. If not provided or None, a new
+                scene is created.
+        """
+        if backend is None and scene is None:
+            raise ValueError("Either backend or scene must be specified.")
+
+        try:
+            import importlib
+
+            backend = importlib.import_module("plato.draw.{}".format(backend))
+        except ImportError:
+            raise ImportError(
+                "Backend plato.draw.{} could not be imported.".format(backend)
+            )
+
+        prim = self._plato_primitive(backend)
+
+        if scene is None:
+            scene = backend.Scene([prim])
+        else:
+            scene.add_primitive(prim)
+        return scene
+
 
 class Shape2D(Shape):
     """An abstract representation of a shape in 2 dimensions."""
--- a/coxeter/shapes/base_classes.py
+++ b/coxeter/shapes/base_classes.py
@@ -139,18 +139,22 @@ class Shape(ABC):
     def to_plato_scene(self, backend=None, scene=None, **scene_kwargs):
         r"""Add this shape to a new or existing :class:`plato.draw.Scene`.
 
-        The plato visualization package provides support for several backends.
-        Each backend supports different primitives (geometry objects) and may
-        not support the primitive corresponding to a specific shape class in
+        The plato visualization package provides support for several backends,
+        including matplotlib, fresnel, povray, pythreejs, and vispy. Each
+        backend supports different primitives (geometry objects) and may not
+        support the primitive corresponding to a specific shape class in
         coxeter. Please refer to the `plato documentation
         <https://plato-draw.readthedocs.io/>`__ for more information about
         supported primitives for each backend.
 
         Args:
             backend (str):
-                Name of backend to use from plato. The backend must support
-                the primitive corresponding to this shape. Defaults to None,
-                in which case the ``scene`` argument must be provided.
+                Name of backend to use from plato. The backend must support the
+                primitive corresponding to this shape. Defaults to None, in
+                which case the ``scene`` argument must be provided.  Supported
+                values include ``"matplotlib"``, ``"fresnel"``, ``"povray"``,
+                ``"pythreejs"``, ``"vispy"``, and ``"zdog"``. See plato
+                documentation for more information.
             scene (:class:`plato.draw.Scene`):
                 Scene object to render into. If not provided or None, a new
                 scene is created.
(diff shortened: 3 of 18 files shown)


bdice:
This PR also fixes a docs issue: https://github.com/executablebooks/jupyter-book/issues/1137

coxeter/families/tabulated_shape_family.py
```diff
@@ -1,8 +1,8 @@
 """Define tabulated shape families.
 
 Tabulated shape families are defined by a dictionary of data that can be
-converted into a shape by some well-defined schema. One example is the
-`GSD shape schema <shapes>`, which defines how to translates dictionaries with
+converted into a shape by some well-defined schema. One example is the `GSD
+shape schema <gsd:shapes>`, which defines how to translates dictionaries with
```
bdice, line 5: This is an unrelated improvement to Sphinx reference syntax. I wanted to namespace this so that the GSD "shapes" reference doesn't conflict with the subpackage name "shapes" of coxeter, should we choose to add a sphinx reference for that. (I didn't know how to use sphinx reference namespaces until recently.)

vyasr:
One possible roadblock to replacing the `plot` method with `to_plato_scene` (discussed offline with @bdice) is that `plot` allows users to integrate a shape directly into a matplotlib figure, which is much easier and more familiar for most users. We should look into whether the plato matplotlib backend has a way to accomplish this, perhaps by constructing a scene that points to an existing figure/axis.

vyasr:
> One possible roadblock to replacing the `plot` method with `to_plato_scene` (discussed offline with @bdice) is that `plot` allows users to integrate a shape directly into a matplotlib figure, which is much easier and more familiar for most users. We should look into whether the plato matplotlib backend has a way to accomplish this, perhaps by constructing a scene that points to an existing figure/axis.

In hindsight, I think all that's required here is to add `ax, fig` to the signature to be forwarded to the scene constructor so that users can treat use this solely via matplotlib. Also maybe a `render=True` kwarg so that it automatically draws on the axis. Users who actually want to interact via plato APIs can then pass `render=False` to get the plato scene and work with it directly if they want.

bdice:
Discussed with @vyasr -- we'll just allow forwarded `**kwargs` to the Scene constructor if `scene is None`.

bdice:
@glotzerlab/coxeter-maintainers I updated the PDF attached at the top of this PR description with an example of matplotlib scenes, per @vyasr's request. Should I commit that Jupyter notebook with example scenes to the `tests/` directory?

bdice:
@glotzerlab/coxeter-maintainers This is finished and ready for review. I verified that the new tests of using plato to draw all supported shapes with matplotlib pass on CI, and they are skipped if matplotlib or plato are unavailable.

coxeter/shapes/base_classes.py
```diff
@@ -107,6 +107,48 @@ def plot(self):
     def __str__(self):
         return repr(self)
 
+    def _plato_primitive(self, backend, **kwargs):
+        raise NotImplementedError("This shape does not have a plato primitive.")
```
vyasr, line 111: I'm anticipating that most people will be unfamiliar with plato, so perhaps a more informative error would just be to indicate that plotting is not supported (which can always be forwarded to devs who do know that this means that the `_plato_primitive` method must be implemented.
```suggestion
        raise NotImplementedError("This shape does not support plotting.")
```

coxeter/shapes/base_classes.py
```diff
@@ -107,6 +107,48 @@ def plot(self):
...
+
+    def to_plato_scene(self, backend=None, scene=None, **scene_kwargs):
```
vyasr, line 113: Can we use matplotlib as a default backend? I realize that not every shape is supported, but the vast majority are (I think it's just spheropolyhedra that are not?) and I would expect matplotlib to be the default choice for most people. We can document that this may not be supported for all shapes, and perhaps catch the relevant exceptions below if we know what plato raises when an invalid primitive is attempted.

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -147,3 +153,15 @@ def __repr__(self):
             f"coxeter.shapes.ConvexSpheropolygon(vertices={self.vertices.tolist()}, "
             f"radius={self.radius}, normal={self.polygon.normal.tolist()})"
         )
+
+    def _plato_primitive(self, backend):
+        verts = _align_points_by_normal(
+            self.normal, self.vertices - self.polygon.center
+        )[0]
```
vyasr, line 160: The validity of this is impacted by #149, right? Will polygons not in the plane be plottable?

coxeter/shapes/polygon.py
```diff
@@ -707,3 +707,12 @@ def __repr__(self):
             f"coxeter.shapes.Polygon(vertices={self.vertices.tolist()}, "
             f"normal={self.normal.tolist()})"
         )
+
+    def _plato_primitive(self, backend):
+        verts = _align_points_by_normal(self.normal, self.vertices - self.center)[0]
```
vyasr, line 712: Same question as for spheropolygons, how does #149 impact this?

coxeter/shapes/polyhedron.py
```diff
@@ -855,3 +855,13 @@ def __repr__(self):
             f"coxeter.shapes.Polyhedron(vertices={self.vertices.tolist()}, "
             f"faces={np.asarray(self.faces).tolist()})"
         )
+
+    def _plato_primitive(self, backend):
+        return backend.Mesh(
+            positions=np.array([self.center]),
+            orientations=np.array([[1.0, 0.0, 0.0, 0.0]]),
+            colors=np.array([[0.5, 0.5, 0.5, 1]] * len(self.vertices)),
```
vyasr, line 863: This makes me wonder, should we support primitive kwargs in addition to scene kwargs so that the user can specify things like the color? I'm also fine with punting on this, just thought I'd ask.

tests/test_plato.py
```diff
@@ -0,0 +1,92 @@
+import tempfile
+
+import pytest
+
+import coxeter
```
vyasr, line 5: This spacing seems strange, why is there a newline between tempfile and pytest?

vyasr: Tests pass, so it's fine, I'm just confused.

vyasr (changes requested):
Looks good. I've requested a few things, but depending on what you think you may not have to do anything for most of them. Ping me once you've addressed them one way or another.

tests/test_plato.py
bdice: isort sections: `FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER` [(source)](https://pycqa.github.io/isort/#custom-sections-and-ordering)

`tempfile` is STDLIB
`pytest` is THIRDPARTY
`coxeter` is FIRSTPARTY.

coxeter/shapes/polyhedron.py
bdice: Good idea, I considered it but I didn't want to bother with that yet. I vote to punt on this.

coxeter/shapes/polygon.py
bdice: This code will need to be updated, yes. It will only support in-plane shapes (because that's what plato supports).

coxeter/shapes/convex_spheropolygon.py
bdice: This code will need to be updated, yes. It will only support in-plane shapes (because that's what plato supports).

coxeter/shapes/base_classes.py
bdice: Sure, we can set a default backend. I don't want to handle catching/re-raising exceptions here. I would prefer to just make a note in the docs and let users handle that.

coxeter/shapes/base_classes.py
bdice: 1. This is only raised when users call a method named `to_plato_scene`, so I think the existing error message here is more descriptive/helpful. Users should not be surprised to see the name "plato."
2. I think "does not support plotting" falsely indicates a deficiency in coxeter rather than the external library (plato) that coxeter is calling.

coxeter/shapes/base_classes.py
```diff
@@ -107,6 +107,48 @@ def plot(self):
...
+        r"""Add this shape to a new or existing :class:`plato.draw.Scene`.
+
+        Args:
+            backend (str):
+                Name of backend to use from plato. The backend must support
+                the primitive corresponding to this shape. Defaults to None,
+                in which case the ``scene`` argument must be provided.
```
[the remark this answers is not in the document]
bdice: Yes, that's a great idea. I will add docs and links to indicate that the matrix of backends/primitives is not fully populated and where to get more info.

tests/test_plato.py
vyasr: I forgot that tempfile and pytest are separate groups, thanks.

coxeter/shapes/polyhedron.py
vyasr: That's fine with me, but if we're punting (rather than making this as `wontfix`) should we instead make `to_plato_scene` accept a dict of `scene_kwargs` rather than forward all `**scene_kwargs` so that we have the flexibility in the future to add a `primitive_kwargs` dict argument or similar?

coxeter/shapes/polygon.py
vyasr: Do you just want to merge this as is for now, then deal with this again later?

coxeter/shapes/base_classes.py
vyasr: That seems like a reasonable compromise to me.

coxeter/shapes/base_classes.py
vyasr: Yes, that's a good point. It does make me wonder whether we should alias `to_plato_scene` with something like `plot` to make it easier for unfamiliar users who might not know what to look for if they don't find something that is actually called `plot`. I'm OK doing that in a separate PR if you'd like to just be done with this one for now, but I do think that would be useful for anyone (especially outside our group) who tries to use it.

coxeter/shapes/base_classes.py
```diff
@@ -107,6 +107,48 @@ def plot(self):
...
+            :class:`plato.draw.Scene`:
+                A scene containing this shape.
+        """
+        if backend is None and scene is None:
+            raise ValueError("Either backend or scene must be specified.")
+
+        try:
+            import importlib
+
+            backend = importlib.import_module("plato.draw.{}".format(backend))
```
bdice, line 138: Note to self: this shouldn't be executed if the scene already exists and has a backend.

coxeter/shapes/base_classes.py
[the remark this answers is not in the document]
bdice: Resolved in d9c74f7b386edc783b06464e592b3375fa2ec39e.

coxeter/shapes/base_classes.py
bdice: I think we can leave it as `to_plato_scene` for now. I think this is a case where explicit is better than implicit. Plotting is very generic, but this functionality is fairly specific to plato. I could see `plot` being an alias for creating a matplotlib figure, for instance.

coxeter/shapes/base_classes.py
bdice: Default backend set to matplotlib in 6f314bd31c83895fc6bbe5e3502f08192343b6f5.

coxeter/shapes/polygon.py
bdice: I agree we should merge this PR and then focus on resolving the conflict with #149 afterward. I created a note on #149 to handle this.

coxeter/shapes/polyhedron.py
bdice: Changed from `**scene_kwargs` to `scene_kwargs` in a3e63e5f0a3c2270e918b7ec841630a980cf8981.

bdice:
@vyasr @Tobias-Dwyer This is ready for another round of review. I will need to make some changes to #149 after this PR is merged and have noted those changes in the description of #149.

coxeter/shape_getters.py
```diff
@@ -24,12 +24,12 @@
...
-    See :ref:`here <shapes>` for the specification of the schema. Note that the
-    schema does not differentiate between 2D and 3D shapes for spheres (vs
-    circles) and ellipsoids (vs ellipses) because in context those can be
-    inferred from simulation boxes.  To address this ambiguity, this function
-    accepts a dimensions parameter that can be used to disambiguate explicitly
-    between these two cases.
+    See :ref:`here <gsd:shapes>` for the specification of the schema. Note
+    that the schema does not differentiate between 2D and 3D shapes for
+    spheres (vs. circles) and ellipsoids (vs. ellipses) because in context
+    those can be inferred from simulation boxes. To address this ambiguity,
```
Tobias-Dwyer, line 30: Change 'those' to 'the dimension of the shape' to eliminate some ambiguity in the writing.

Tobias-Dwyer (commented):
just a small comment to disambiguate the language of the documentation and you should be good!

coxeter/shape_getters.py
bdice: Good idea. Done!

vyasr:
I still owe a review here, just letting you know that it's on my list.

coxeter/shape_getters.py
```diff
@@ -24,12 +24,12 @@
...
-    circles) and ellipsoids (vs ellipses) because in context those can be
-    inferred from simulation boxes.  To address this ambiguity, this function
-    accepts a dimensions parameter that can be used to disambiguate explicitly
-    between these two cases.
+    See :ref:`here <gsd:shapes>` for the specification of the schema. Note that
+    the schema does not differentiate between 2D and 3D shapes for spheres (vs.
+    circles) and ellipsoids (vs. ellipses) because in context the
+    dimensionality of the shape can be inferred from simulation boxes. To
+    address this ambiguity, this function accepts a dimensions parameter that
+    can be used to disambiguate explicitly between these two cases.
```
vyasr, line 32:
```suggestion
    circles) and ellipsoids (vs. ellipses) because in context the
    dimensionality of those shapes can be inferred from simulation boxes. To
    address this ambiguity, this function accepts a ``dimensions`` parameter
    that can be used to disambiguate explicitly between these two cases.
```

coxeter/shapes/base_classes.py
```diff
@@ -133,6 +133,66 @@ def plot(self):
...
+                Keyword arguments forwarded to the :class:`plato.draw.Scene`
+                (Default value: None). Only used if ``scene`` is not provided
+                or None.
+
+        Returns:
+            :class:`plato.draw.Scene`:
+                A scene containing this shape.
+
+        Raises:
+            `NotImplementedError`:
```
vyasr, line 172:
```suggestion
            NotImplementedError:
```

coxeter/shapes/base_classes.py
```diff
@@ -133,6 +133,66 @@ def plot(self):
...
+                If no plato primitive corresponds to this coxeter shape class.
+            `AttributeError`:
```
vyasr, line 174:
```suggestion
            AttributeError:
```

vyasr (approved):
LGTM, thanks!

PR #133, opened 2020-12-19
Use public properties everywhere.

bdice:
## Description
This PR uses public properties everywhere possible.

## Motivation and Context
Public properties are validated when set and their names are shorter. All attributes that can be set should call the property setter in `__init__` instead of duplicating the validation logic in the constructor.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:


bdice:
The failing docs are fixed in PR #132. We can cherry-pick that change onto `master` if that PR is delayed.

vyasr (approved):
LGTM

PR #122, opened 2020-10-13
perimeter now accepts a radius of zero in convexspheropolygon

Tobias-Dwyer:
perimeter now accepts a radius of zero in convexspheropolygon
## Description

perimeter now accepts a radius of zero in convexspheropolygon
## Motivation and Context

Circle doesn't accept a circle of zero therefore the previous code produced an error for a radius of 0. 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/coxeter/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/coxeter/blob/master/ContributorAgreement.md).
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/coxeter/blob/master/ChangeLog.txt) and the [credits](https://github.com/glotzerlab/coxeter/blob/master/doc/source/credits.rst).

PR #138, opened 2021-01-12
Rewrite tests to explicit reuse fixtures to avoid hypothesis errors.

vyasr:
## Description

This PR rewrites test that were implicitly reusing pytest fixtures to make this reuse explicit. I added a changelog entry since this may be important for others to track the reason for the failures.

## Motivation and Context

#122 added various tests that reuse pytest fixtures in hypothesis tests. Those tests would pass with warnings in older versions of hypothesis, but the [recent release of hypothesis 6.0.0 makes this an error](https://hypothesis.readthedocs.io/en/latest/changes.html#v6-0-0). These changes will avoid problems in the future.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -9,6 +9,12 @@ Added
 
 - Ellipse area setter and Ellipsoid volume setter.
 
+Changed
+~~~~~~~
+
+- Ensure that hypothesis-based tests don't implicitly reuse pytest fixtures.
+
+
 v0.4.0 - 2020-10-14
 -------------------
 
--- a/tests/test_spheropolygon.py
+++ b/tests/test_spheropolygon.py
@@ -37,31 +37,43 @@ def test_2d_verts(square_points):
     ConvexSpheropolygon(square_points, 1)
 
 
-@given(r=floats(0.1, 1000))
-def test_radius_getter_setter(square_points, r):
+def test_radius_getter_setter(square_points):
     """Test getting and setting the radius."""
-    square_points = square_points[:, :2]
-    convexspheropolygon = ConvexSpheropolygon(square_points, r)
-    assert convexspheropolygon.radius == r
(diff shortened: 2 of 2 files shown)


bdice:
@Tobias-Dwyer Just so you're aware as a maintainer, we usually catch problems introduced by testing dependency upgrades by using a tool called dependabot for freud and signac. It creates PRs like [this one](https://github.com/glotzerlab/freud/pull/710). It could be useful to set up for coxeter. We would pin all testing dependencies in `requirements-testing.txt` and let dependabot upgrade that on a monthly basis.

bdice (changes requested):
@vyasr We should just use
```python
@pytest.fixture(scope="module")
```
instead.

vyasr:
@bdice why? While that makes the error go away, that assumes that any change in the fixture from one test is safe to persist to another, right? That _might_ be true, but I don't see why you would want to assume that when so many tests modify the fixture. Even if that's currently true, it would be easy to see adding one future tests that makes a nontrivial modification (e.g. changing the vertices of a shape since we've discussed making vertices settable) in a way that violates assumptions in another test.

bdice:
> @bdice why? While that makes the error go away, that assumes that any change in the fixture from one test is safe to persist to another. That _might_ be true, but I don't see why you would want to assume that when so many tests modify the fixture. Even if that's currently true, it would be easy to see adding one future tests that makes a nontrivial modification (e.g. changing the vertices of a shape since we've discussed making vertices settable) in a way that violates assumptions in another test.

I discussed with @vyasr offline and retract my suggestion. We wouldn’t want to allow tests to modify the value of the fixture and mess up other tests.

bdice (approved):
Looks good to me. I'll merge this so that the other approved PRs can be updated and merged.

Issue #129, opened 2020-11-11
"center" should mean Centroid (Center of Mass for shapes with uniform density)


bdice:
Currently the `polyhedron.center` property appears to be defined as an average of the vertices. However, this does not always coincide with the shape's center of mass: https://bell0bytes.eu/centroid-convex/

https://github.com/glotzerlab/coxeter/blob/0b02755cb0cf9550f1ed3a94b5c9eeed21d25830/coxeter/shapes/polyhedron.py#L474-L477

I recommend the following:
- Clarify and define relevant properties (`center` = center of mass = centroid = barycenter) <sup>[1]</sup>
- The current behavior of `center` could be retained under another name, but the proper name for this is unclear to me

<sup>[1]</sup> I note that [Wikipedia says](https://en.wikipedia.org/wiki/Centroid) the center of mass, centroid, and barycenter are synonymous for shapes of uniform density:
> While in geometry the word barycenter is a synonym for centroid, in astrophysics and astronomy, the barycenter is the center of mass of two or more bodies that orbit each other. In physics, the center of mass is the arithmetic mean of all points weighted by the local density or specific weight. If a physical object has uniform density, its center of mass is the same as the centroid of its shape.

vyasr:
Here are my thoughts
- The reason for the Wikipedia statement is that the centroid is the average of all _points_ in the figure, where points does not mean _vertices_ but literally all points. That would mean performing an integral and then dividing by the area, which is the calculation I was referring to in the discussion today when I said that for a constant density the two should coincide. I conveniently forgot that averaging over all points is in no way equal to averaging over the vertices.
- I don't think there is any issue with redefining the current behavior of center, although we should probably verify that before deciding whether to change it. In any case I would prefer to define a new property `com` or `center_of_mass` and move the existing definition of `center` to some more appropriate name. The final thing to do would be to decide which one `center` should be an alias for; I would also lean towards center of mass, but I think that should be assessed more carefully.
- Properties like moments of inertia etc should still be calculated relative to the origin of the coordinate system. We have discussed this before, and there are various reasons to prefer that over implicitly using the center of mass. If a user wants properties computed relative to the center of mass, they should set the com to the origin.

bdice:
[...]

PR #141, opened 2021-01-15
2D shape kernels (or distance to surface)

Tobias-Dwyer:
Shape kernels for all 2D shapes in coxeter except for general polygons. 

## Description

Shape kernels are useful because they are used in several theories and provide a detailed way to characterize a shape. Here we include them for everything in 2D except for general polygons. 

## Motivation and Context

New feature, we didn't have shape kernels before. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


coxeter/shapes/circle.py
```diff
@@ -159,3 +159,19 @@ def iq(self):
...
+
+        Args:
+            value (array):
+                Points over which to calculate the shape kernel
+                which can only be from negative 0 to 2pi.
+
+        Returns:
+            kernel
+        """
+        return value * 0.0 + self.radius
```
vyasr, line 177:
```suggestion
        return np.ones_like(value) * self.radius
```

coxeter/shapes/circle.py
```diff
@@ -159,3 +159,19 @@ def iq(self):
...
         return 1
+
+    def shape_kernel(self, value):
+        """Shape kernel from 0 to 2pi.
+
+        Kernel is simply the radius for a circle so we
+        simply replace array elements with the radius.
+
+        Args:
+            value (array):
```
vyasr, line 170: I would double-check how we document array arguments. In theory this could also be a list, tuple, etc. Usually those are called "array-like" objects by numpy, but we may not be consistent about that in coxeter.

coxeter/shapes/circle.py
```diff
@@ -159,3 +159,19 @@ def iq(self):
...
+                Points over which to calculate the shape kernel
+                which can only be from negative 0 to 2pi.
+
+        Returns:
+            kernel
```
vyasr, line 175: Include the type of the return and discuss its shape. For example, something like `np.ndarray: An array of distances of the same shape as the input`.

coxeter/shapes/circle.py
```diff
@@ -159,3 +159,19 @@ def iq(self):
         This is 1 by definition for circles.
         """
         return 1
+
+    def shape_kernel(self, value):
```
vyasr, line 163: Use a better argument name than "value". Are these angles? Is that the same for all 2D shapes? For 3D, will it be a pair of angles? It's worth thinking about the 3D case so that you can make the function arguments look the same, at least for the comment one. Maybe literally theta/phi for spherical coordinates?

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
 
         radius = np.min(distances)
         return Circle(radius, self.center)
+
+    def shape_kernel(self, value):
+        """Shape kernel from 0 to 2pi.
+
+        This algorithm assumes the vertices are ordered
+        counterclockwise and that they start in the first quadrant.
```
vyasr, line 134: The counterclockwise ordering is true of polygon shapes in coxeter by construction. Can we relax the "starting in the first quadrant assumption"?

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
...
+        pairs = []
+        slopes = []
+        y_int = []
+        angle_range = np.zeros((len(verts), 2))
+
+        for i in range(len(verts)):
+            if i < len(verts) - 1:
+                pairs.append([verts[i, :], verts[i + 1, :]])
+            else:
+                pairs.append([verts[i, :], verts[0, :]])
```
vyasr, line 159: Have a look a [`numpy.roll`](https://numpy.org/doc/stable/reference/generated/numpy.roll.html). What you are doing should be equivalent to `pairs = list(zip(verts, np.roll(verts, axis=0, shift=-1)))`

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
...
+                Points over which to calculate the shape kernel
+                which can only be from negative 0 to 2pi.
+
+        Returns:
+            kernel
+        """
+        verts = self.vertices[:, :2]
+        theta = value
+        verts[:, 0] = verts[:, 0] - np.average(verts[:, 0])
+        verts[:, 1] = verts[:, 1] - np.average(verts[:, 1])
```
vyasr, line 147: A few things here:
1. You can pass `np.average` an axis, so these two lines are equivalent to `verts -= np.average(verts, axis=0)`.
2. This average is equivalent to the center as currently defined, so you could also write `verts = self.vertices[:, :2] - self.center[:, :2]`.
3. In light of #129, we might want the kernel function to take an argument that allows us to pass the "center". It could default to `self.center`, but in case it makes sense to use a different center (e.g. the center of mass rather than just the average of the vertices) you could provide it.

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
...
+        slopes = []
+        y_int = []
+        angle_range = np.zeros((len(verts), 2))
+
+        for i in range(len(verts)):
+            if i < len(verts) - 1:
+                pairs.append([verts[i, :], verts[i + 1, :]])
+            else:
+                pairs.append([verts[i, :], verts[0, :]])
+        for i in range(len(pairs)):
```
vyasr, line 160: You can replace almost all of the for loops in this code using numpy functions that will do the looping internally (and faster) for you. @tommy-waltmann would you mind helping @Tobias-Dwyer with that?

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
...
+                theta_initial = max_angle
+                angle_range[i, 0] = min_angle
+                angle_range[i, 1] = max_angle
+            else:
+                angle_range[i, 0] = max_angle
+                angle_range[i, 1] = 2 * np.pi
+
+        # Make the kernel:
+
+        kernel = 0 * theta
```
vyasr, line 202: Usually better to be more explicit with assignments like this: `kernel = np.zeros_like(theta)` or `kernel=np.zeros(theta.shape)`.

coxeter/shapes/convex_polygon.py
```diff
@@ -126,3 +126,121 @@ def incircle_from_center(self):
...
+        angle_range = angle_range
+
```
vyasr, line 204:
```suggestion

```

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -187,3 +187,107 @@ def perimeter(self, value):
...
+
+            # define the angle range for rounding
+            pt1 = v2 + v12n * self.radius
+            pt3 = v2 + v32n * self.radius
+            angle_ranges.append(
+                (self._get_polar_angle(pt1), self._get_polar_angle(pt3))
+            )
+
+        # compute shape kernel for the new set of vertices
+        kernel = ConvexPolygon(new_verts).shape_kernel(value)
```
vyasr, line 274: Why do you need to make a new polygon? Won't `self._polygon.shape_kernel(value)` work just as well (since it'll internally center the shape)?

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -187,3 +187,107 @@ def perimeter(self, value):
...
+
+        # compute intermediates
+        for i in range(num_verts):
+            v1 = verts[i - 1]
+            v2 = verts[i]
+            v3 = verts[i + 1] if i + 1 < num_verts else verts[0]
+            v12 = v1 - v2
+            v32 = v3 - v2
+            v12n = self._get_outward_unit_normal(v12, v2)
+            v32n = self._get_outward_unit_normal(v32, v2)
```
vyasr, line 256: At least up to this point it should be able to rewrite the for loop using `numpy.roll` followed by some straightforward array subtractions. It also avoids the confusing asymmetry where you have to manually wrap `i+1` but not `i-1` because negative indices work.

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -187,3 +187,107 @@ def perimeter(self, value):
...
+            angle_ranges.append(
+                (self._get_polar_angle(pt1), self._get_polar_angle(pt3))
+            )
+
+        # compute shape kernel for the new set of vertices
+        kernel = ConvexPolygon(new_verts).shape_kernel(value)
+
+        # get the shape kernel for this shape by adjusting indices of shape kernel
+        # for the new vertices
+        new_kernel = kernel
```
vyasr, line 278: No need for a new variable, you can just keep modifying the kernel directly.

tests/test_circle.py
```diff
@@ -109,3 +110,20 @@ def test_invalid_radius_setter():
     circle = Circle(1)
     with pytest.raises(ValueError):
         circle.radius = -1
+
+
+@given(floats(0.1, 10))
+def test_shape_kernel(r):
```
vyasr, line 116: I like this test. It's almost identical for all the different cases though, is it worth moving it to a helper function in conftest.py that generates the xy array, transposes it, makes the hull, then checks the equality of a set of properties (could write those explicitly, or generalize by using `getattr(obj, x)` for a provided set of strings as properties).

vyasr (changes requested):
This looks good, but not quite ready to merge yet. Most of the comments that I made in any one file are applicable to multiple files, so try and apply the appropriate requests everywhere. Also @tommy-waltmann thoughts on a name for the property? I think something like `distance_to_surface` seems more appropriate.

coxeter/shapes/convex_spheropolygon.py
tommy-waltmann: The new vertices are different than the vertices of the polygon held internally by the `Spheropolygon` object. The new vertices are that of the old polygon, but with a buffer of size `rounding_radius` added. This is hard to explain here, If you need more clarification I can provide it.

coxeter/shapes/circle.py
tommy-waltmann: I think `angles` is a good name. In 2D, this is just a list of values between 0 and 2pi, but in 3D this would be a list of pairs (theta, phi) defined as `0<=theta <=pi` and `0<=phi<=2pi`.

coxeter/shapes/convex_spheropolygon.py
vyasr: Ah I missed this, I though it was just being centered. Never mind then. Am I correct to think that here you are calculating the kernel for a new polygon where all edges are pushed outward by the rounding radius, and that the loop below then subtracts the difference between this polygon and the true spheropolygon wherever that difference is nonzero (the `angle_ranges` variable)?

coxeter/shapes/circle.py
vyasr: Yeah, so in 3D theta/phi is probably the right answer. In 2D either angle or theta would be fine IMO.

vyasr:
Please re-request my review once my requests are addressed, or when you have any questions about them. Thanks!

coxeter/shapes/convex_spheropolygon.py
tommy-waltmann: You are mostly right. It computes the shape kernel for the polygon with the edges pushed out by the rounding radius. In the second step, it doesn't really subtract, it just computes what the distance should be in the rounded sections using the law of cosines and variables in the problem.

tommy-waltmann:
@vyasr In fixing the documentation, would you suggest having a more detailed description of what the shape kernel is in an abstract method in `Shape2D`, and then having shorter descriptions in the documentation for each implementing class?

vyasr:
@tommy-waltmann a couple of thoughts
- I like the idea of putting a method in the base classes, but I don't think it should be abstract. Instead put it in `Shape` (not `Shape2D`) and have it raise a `NotImplementedError`. The difference is that this won't prohibit child classes from being instantiated the way that an abstract method would. Not every class _has_ to implement this method, the ones that don't just need to fail when someone tries to use it.
- As I've mentioned before (and is now in the PR title), I would prefer to rename this method to `distance_to_surface` or something like that which is much easier to understand than `shape_kernel`. That should mostly take care of the documentation issue since the name is self-explanatory.

coxeter/shapes/convex_spheropolygon.py
tommy-waltmann: I agree that `np.roll` would be useful in resolving the asymmetry, although I don't know that there would be much performance payoff unless the entire for loop was vectorized, and I'm not sure how I would vectorize the `_get_outward_unit_normal` helper function, simply because of all the if-else divergences that are necessary to avoid dividing by 0.

tommy-waltmann: I'll try to vectorize as much of this as I can.

vyasr: I think you should be able to vectorize that `_get_outward_unit_normal` method using some fancy numpy indexing, but I definitely agree that that function will be harder to handle than everything else. I would suggest cleaning up the rest of this code before coming back to this function, and assuming all of that works I can help you with this function as needed.

coxeter/shapes/convex_polygon.py
Tobias-Dwyer: I implemented the relaxation. You can now start in any quadrant you would like.

coxeter/shapes/convex_polygon.py
Tobias-Dwyer: I've left this as the average for now but from your above comment you want to have the center of the shape be what we use rather than the average. For convex shapes is the average of the verts always the center? (this I think is what comment 2 is saying, just wanna make sure it's true for all cases.)

coxeter/shapes/convex_polygon.py
Tobias-Dwyer: this has been added.

coxeter/shapes/convex_polygon.py
vyasr: What I'm suggesting is replacing these two lines with `verts = self.vertices - self.center`, which would be much more concise and also make it much easier in the future to change it to `verts = self.vertices - SOMETHING_ELSE`, where the something else could be e.g. the center of mass instead of the average of the vertices.

vyasr:
@Tobias-Dwyer @tommy-waltmann I made some minor cleaning changes in my last commit, mainly to avoid using `inf` as a string when numpy has a way to represent it. I'm going to try a different approach for this distance and see how it works.

vyasr:
@Tobias-Dwyer @tommy-waltmann I did some cleanup of the convex_polygon code and made it a bit faster (not a huge change, maybe ~20%). Have a look at it and see if there are similar changes to be made to the spheropolygon code. I also made the code support out of plane polygons, but since the rotation into the plane is ambiguous (rotation about the z axis is an uncontrolled degree of freedom) the angle must be defined in an unambiguous way relative to that rotation. We can either try and resolve this in a determinate way, or we can just throw a warning (or an error) if a user tries to use this method on a shape that's rotate out of the plane. What do you think?

One other question, should we add an argument `from` that can be used to specify the location from which the distances are computed. Right now it default to `self.center`, but it might also make sense to use (for example) the center of the smallest bounded sphere (which may not be the center of the shape). We could give a set of valid choices ("center", "insphere_center", etc), or we could allow an arbitrary 3-vector and check that it falls inside the shape. Thoughts on whether this is useful/something we should do?

Tobias-Dwyer:
@vyasr I think simply throwing an error if things are rotated out of plane would be best. Just because for most use cases I don't think users are going to feed in xyz vertices for 2D that is rotated out of plane. As far as specifying the center that the user wants, I'm not super sure of the practical use of this off the top of my head, though @tommy-waltmann please jump in if this would help your work. Rn I would just go with a single center, though the code would move fine if we let users define an arbitrary point within the shape, it might add some confusion in the future if a shape with an odd center is defined.

vyasr:
> @vyasr I think simply throwing an error if things are rotated out of plane would be best. Just because for most use cases I don't think users are going to feed in xyz vertices for 2D that is rotated out of plane. As far as specifying the center that the user wants, I'm not super sure of the practical use of this off the top of my head, though @tommy-waltmann please jump in if this would help your work. Rn I would just go with a single center, though the code would move fine if we let users define an arbitrary point within the shape, it might add some confusion in the future if a shape with an odd center is defined.

I'm fine with throwing the error for out-of-plane polygons. Regarding the center, here's the immediately obvious case: if the average of the vertices (the current definition of self.center) does not coincide with the centroid, which would be the natural choice? Is it always the same? What if you have a nonconvex polygon?

Tobias-Dwyer:
> > @vyasr I think simply throwing an error if things are rotated out of plane would be best. Just because for most use cases I don't think users are going to feed in xyz vertices for 2D that is rotated out of plane. As far as specifying the center that the user wants, I'm not super sure of the practical use of this off the top of my head, though @tommy-waltmann please jump in if this would help your work. Rn I would just go with a single center, though the code would move fine if we let users define an arbitrary point within the shape, it might add some confusion in the future if a shape with an odd center is defined.
> 
> I'm fine with throwing the error for out-of-plane polygons. Regarding the center, here's the immediately obvious case: if the average of the vertices (the current definition of self.center) does not coincide with the centroid, which would be the natural choice? Is it always the same? What if you have a nonconvex polygon?

Well for the applications of the shape kernel for most applications (as you know @vyasr @tommy-waltmann just repeating here for clarity) the shape kernel can be thought of as a perturbation from a sphere or circle. So I would say for a shape with many vertices we should just define the center of the kernel as the center of the minimum circumcircle. This answer would work for concave shapes but of course only if a shape kernel exists, which I haven't thought about how to check for concave shapes.

vyasr:
Assuming you mean incircle (not circumcircle), I think that's mostly right, but that still leaves possible ambiguity. For a trivial example, consider a rectangle where l < w. You could theoretically choose any number of circles with radius r=l by sliding along the center line. In this case there is a canonically correct choice (set the center at w/2), but for more complex shapes that may not be obvious. You don't plan to define this method for general polygons, right? Only for convex polygons? In which case, is it always true that the center you want is the center of the largest _concentric_ sphere that fits in the shape? If so, it might be safe to just assume that, and we can add a `from` feature later. However, the current implementation (I didn't change this) is relative to `self.center`, but it really should be relative to `self.some_insphere.center` (where we can pull `some_insphere` from one of the properties added in #145).

tommy-waltmann:
I'm also fine with throwing an error if the shape is out of the plane. I like the idea of making the distances relative to the center of the largest concentric circle that fits inside the shape. I can't think of a use case for different centers right now, but I imagine that I could run into one later. Let's worry about that later, I don't think it would require a lot of work to implement anyway.

vyasr:
I think that's fine for now. I just merged in master, so all the different type of insphere calculations are now available. The one that you want is `maximal_centered_bounded_[circle|sphere]` (2D vs 3D). I would also mention in the docstring that this is what you're using.

Tobias-Dwyer (commented):
looks good, like the additional numpy logic added!

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -142,6 +147,110 @@ def perimeter(self, value):
...
+        new_verts = v2 + uvec * self.radius / (np.sin(phi / 2)[:, None])
+
+        # define the angle range for rounding
+        pt1 = v2 + v12n * self.radius
+        pt3 = v2 + v32n * self.radius
+
+        angle_ranges = [
+            (self._get_polar_angle(pt1[i]), self._get_polar_angle(pt3[i]))
+            for i in range(num_verts)
+        ]
```
tommy-waltmann, line 231: Out of curiosity, @vyasr did you find a way we might be able to vectorize this operation or the `_get_outward_unit_normal` operation as you went through this code?

tommy-waltmann: I think perhaps `_get_polar_angle` could be vectorized by making use of `np.arctan2` in some fashion.

tommy-waltmann (commented):
A small change or two that I can handle, and a quick question.

coxeter/shapes/convex_spheropolygon.py
vyasr: Oh I missed this one. Yes, this should be possible to vectorize by basically inlining the logic from `_get_polar_angle`. Let me know if you'd like me to take a stab at that.

vyasr: This is now using `np.arctan2`.

tommy-waltmann (approved):

PR #145, opened 2021-01-28
Feature/inspheres

vyasr:
## Description

Proper inspheres/circles are added for polyhedra and polygons, and the currently present properties for largest contained spheres are renamed. A generic set of parent properties are added for the largest contained sphere and setting its radius, and the properties are implemented for all shapes where it is possible.

Note on naming: the word "insphere" is highly overloaded and extremely inconsistently used in the literature and in a few geometry textbooks that I've looked at. There is frequent disagreement on 1) whether it should be concentric and 2) what conditions must be imposed regarding its relation to polytope faces, when referring to the insphere (incircle) of a polyhedron (polygon). The closest thing to consensus in the geometry literature is regarding the definition of an inscribed sphere (circle) of a polyhedron (polygon) as the sphere contained in the shape that is tangent to all faces (edges). As a result, I've chosen to exclusively use the term insphere (incircle) for this property of polyhedra/polygons, which may or may not exist for any particular shape. For all other shapes, in the interest of avoiding any ambiguity, I introduced a new property `maximal_bounded_sphere` as a direct analogue to the `minimal_bounding_sphere`.

I am open to alternative choices of naming, with the following caveats:
1) I will likely reject any suggestion involving "insphere" out of hand due to the previously mentioned ambiguities and the fact that it will introduce cognitive dissonance with the definitions in the Polygon and Polyhedron classes.
2) I specifically chose "maximal_bounded" rather than "largest_bounded" sphere despite the latter sounding a bit more natural to me in order to reinforce the clear relationship to the "minimal_bounding_sphere", which _is_ accepted terminology. I could be convinced to change this, but I would probably need a more convincing argument than "it sounds strange" to outweigh the didactic benefits of the current naming.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


Tobias-Dwyer (approved):
Didn't see any errors, and the terminology looks consistent.

PR #149, opened 2021-02-02
Require polygons to be in xy plane for some computations.

bdice:
## Description
This introduces the private method `_requires_xy_plane()` to Polygon and ConvexSpheropolygon.

I also fixed GSD shape specifications for Polygons and ConvexSpheropolygon so they properly adhere to the spec, which requires vertices to be of shape `(N, 2)`. See: https://gsd.readthedocs.io/en/stable/shapes.html#polygons

To-do:
- [ ] Require 2D shapes to be in-plane, once #132 is merged. ([link 1](https://github.com/glotzerlab/coxeter/pull/132#discussion_r571506288), [link 2](https://github.com/glotzerlab/coxeter/pull/132#discussion_r571506422))

## Motivation and Context
Enforcing this requirement ensures that the shape properties make sense.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/coxeter/shapes/convex_spheropolygon.py
+++ b/coxeter/shapes/convex_spheropolygon.py
@@ -52,6 +52,18 @@ class ConvexSpheropolygon(Shape2D):
         if not _is_convex(self.vertices, self._polygon.normal):
             raise ValueError("The vertices do not define a convex polygon.")
 
+    def _require_xy_plane(self, allow_negative_z=False):
+        normal = self.normal
+        if allow_negative_z:
+            normal = np.abs(normal)
+        if not np.array_equal(normal, np.array([0, 0, 1])):
+            class_name = type(self).__name__
+            raise ValueError(
+                f"This method requires the {class_name} to be embedded in the xy "
+                "plane with a normal vector pointing along the positive z "
+                f"direction. The normal of this {class_name} is {self.normal}."
+            )
+
     @property
     def polygon(self):
         """:class:`~coxeter.shapes.ConvexPolygon`: The underlying polygon."""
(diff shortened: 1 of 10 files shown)


coxeter/shapes/convex_spheropolygon.py
```diff
@@ -29,20 +29,20 @@ class ConvexSpheropolygon(Shape2D):
 
     Example:
         >>> rounded_tri = coxeter.shapes.ConvexSpheropolygon(
-        ...   [[-1, 0], [0, 1], [1, 0]], radius=.1)
+        ...   [[1, 0], [0, 1], [-1, 0]], radius=.1)
```
bdice, line 32: @glotzerlab/coxeter-maintainers I switched this test (and a few others) so that they find the normal vector to be `[0, 0, 1]` instead of `[0, 0, -1]`. I'm not sure if we ought to do something to change behavior for re-ordering of clockwise-oriented points. I could enable `allow_negative_z=True` in the `_requires_xy_plane` function for cases where it makes sense to allow clockwise orientations.

vyasr:
@bdice are there unresolved questions you have regarding this PR, or is it just waiting for you to find time for it? If the latter, would it help to work with @Tobias-Dwyer on it?

bdice:
@vyasr Just needs dev time. If @Tobias-Dwyer is interested in working on it, that might help, but it would take some re-explanation of the behavior changes that you and I discussed at length. I'll get back to this PR at some point, I just need to set aside an uninterrupted block of time since it requires a lot of changes in tandem.

bdice:
Closing as stale.

PR #150, opened 2021-02-02
Remove center and bounding/bounded spheres/circles from sphero-shapes.

bdice:
## Description
Removes `center`, `centroid`, and bounding circles/spheres from the spheropolygon and spheropolyhedron classes. The implementation of `centroid` was not valid for arbitrary sphero-shapes.

## Motivation and Context
If it ain't broke, don't fix it. If it is broke, delete it before shipping. `#cantfixwontfix`

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -12,15 +12,15 @@ Added
 - Point in ellipse checks.
 - Inertia tensors for 2D shapes that implement moments of inertia.
 - Add minimal bounding sphere for all shapes.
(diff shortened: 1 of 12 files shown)


vyasr (approved):
LGTM thanks for fixing! I made the small modification of adding centroid as a property at the `Shape` level and defining `center` as an alias for it there so that we can consistently impose this API choice across all shapes with duplication.

PR #160, opened 2022-02-02
Use pre-commit hooks to run linters and update versions for NEP 29

### LINKED ISSUE #128: Determine minimum required versions of dependencies
Currently numpy and scipy are listed as requirements without any particular version required. We should determine specific minimum version requirements and add them to both setup.py and requirements.txt.

vyasr:
## Description

This is a maintenance PR that performs a few tasks to bring coxeter up-to-date and simplify development. It modernizes the pre-commit config file to use current versions of hooks and stops relying on a local flake8 installation, instead installing flake8 into the pre-commit environment along with all the necessary plugins. This PR also enables pre-commit.ci and removes the style check step of CI entirely to avoid duplicating those runs. Finally, it updates the supported Python/NumPy versions to match the NEP 29 cadence.

## Motivation and Context

This change avoids version-related ambiguity for linters and simplifies the process of running hooks locally by creating a single entrypoint. It also reduces CI time by using pre-commit.ci. NEP 29 adherence simplifies development and maintenance. Resolves #128.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


.circleci/config.yml
```diff
@@ -16,8 +16,9 @@ jobs:
       - run: &style-check
           name: Style check
           command: |
-            python -m pip install --progress-bar off --user -U flake8 flake8-black flake8-bugbear flake8-docstrings flake8-rst-docstrings pep8-naming flake8-isort
-            python -m flake8 --show-source ./
+            python -m pip install --progress-bar off --user -U pre_commit
```
bdice, line 19: The package name is `pre-commit`. It is correct to invoke it as `python -m pre_commit` if you don't want to use the alias `pre-commit`.
```suggestion
            python -m pip install --progress-bar off --user -U pre-commit
```

.circleci/config.yml
```diff
@@ -16,8 +16,9 @@ jobs:
...
+            python -m pre_commit install
```
bdice, line 20: This installs git hooks -- is this desired? Probably can skip this command.

.pre-commit-config.yaml
```diff
@@ -16,7 +19,7 @@ repos:
       - id: check-json
       - id: check-yaml
   - repo: https://github.com/pycqa/isort
```
bdice, line 21: Canonical URL:
```suggestion
  - repo: https://github.com/PyCQA/isort
```

.pre-commit-config.yaml
```diff
@@ -25,15 +28,24 @@ repos:
...
+    rev: '6.1.1'
     hooks:
-      - id: flake8
-        name: flake8
-        entry: flake8
-        language: python
-        types: [python]
+      - id: pydocstyle
+        exclude: '(?:extern/.*|tests/.*|setup.py|conf.py)'
+  - repo: https://gitlab.com/pycqa/flake8
```
bdice, line 36: No longer using gitlab:
```suggestion
  - repo: https://github.com/PyCQA/flake8
```

.pre-commit-config.yaml
```diff
@@ -25,15 +28,24 @@ repos:
...
+    rev: '4.0.1'
+    hooks:
+    -   id: flake8
+        additional_dependencies: [
+            'flake8-bugbear==22.1.11',
+            'flake8-rst-docstrings==0.2.5',
+            'pep8==1.7.1',
+            'pep8-naming==0.12.1',
+			'flake8-comprehensions==3.8.0',
```
bdice, line 45: Spaces, not tabs.
```suggestion
            'flake8-comprehensions==3.8.0',
```

.circleci/config.yml
vyasr: PyPI has a redirect for `pre_commit` so that you can use either safely.

bdice: ...but let's use the canonical name. Those redirection rules could change in the future.

vyasr: I mean sure I can change it, but I think that's _highly_ unlikely given how many packages I've seen using `pre_commit` in the past. And redirects aren't exactly hard to maintain.
Regarding using `pre-commit` vs `python -m pre_commit`, I find it preferable to sidestep potential issues with what's on the PYTHONPATH where possible. I could make this even cleaner and less error-prone by using a virtual environment, but I'm punting on that for now.

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ version: 2
 jobs:
   style-check:
     docker:
-      - image: circleci/python:3.9
+      - image: circleci/python:3.10
```
bdice, line 10: Let's try migrating from legacy `circleci` to the newer `cimg` images. If it fails, we can revert that change for this PR - just trying to future-proof the CI a bit.

https://circleci.com/docs/2.0/next-gen-migration-guide/
https://hub.docker.com/r/cimg/python

```suggestion
      - image: cimg/python:3.10
```

.pre-commit-config.yaml
```diff
@@ -25,15 +28,24 @@ repos:
...
+    rev: '6.1.1'
     hooks:
-      - id: flake8
-        name: flake8
-        entry: flake8
-        language: python
-        types: [python]
+      - id: pydocstyle
+        exclude: '(?:extern/.*|tests/.*|setup.py|conf.py)'
+  - repo: https://gitlab.com/PyCQA/flake8
```
bdice, line 36: The repo has moved from GitLab to to GitHub.
```suggestion
  - repo: https://github.com/PyCQA/flake8
```

bdice (approved):

.circleci/config.yml
vyasr: Good call, forgot this shift happened.

.pre-commit-config.yaml
vyasr: Whoops missed this one.

vyasr:
@joaander it occurs to me that the changes to the required Python versions in this PR may be problematic for https://github.com/glotzerlab/software. coxeter claims to be following NEP 29, and this PR brings it up to date, but I recall we had this discussion before where the required numpy versions ended up causing problems due to the versions provided by the Ubuntu package manager. How have you address NEP 29 support for other packages in the container like HOOMD, if at all? If it causes problems, is there any middle ground that you've found, or is the result simply that we can't follow NEP 29 on our packages?

vyasr:
Josh approved this offline.

PR #171, opened 2023-02-06
Added new classes to return the edges of polyhedra

janbridley:
## Description
The new class ``polyhedron.edges`` returns the a list of the edges of a polyhedron as vertex-index pairs, similar to the current ``polyhedron.faces`` class. The class ``polyhedron.get_edge_vectors`` returns a list of edges as vectors in 3D space.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,36 @@ def vertices(self):
...
+    def edges(self):
+        """list(:class:`numpy.ndarray`): Get the polyhedron's edges.
+
+        Results returned as vertex index pairs.
+        """
+        edges = []
+        for face in self.faces:
+            [
+                edges.append((i, j))
+                for i, j in zip(np.append(face, face[0]), np.append(face[1:], face[0]))
```
vyasr, line 375: `np.roll` is the nice way to do this!
```suggestion
                for i, j in zip(face, np.roll(face, -1))
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,36 @@ def vertices(self):
...
+                if (j, i) not in edges
```
vyasr, line 376: I've forgotten how I set this up; are we guaranteed by the handedness requirements of the polyhedron that no two faces will try to add the same (i, j) pair with the same order, so we only have to check the reverse? I believe that is correct, just want to make sure that it's not an oversight here.

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,36 @@ def vertices(self):
...
+    @property
+    def edges(self):
+        """list(:class:`numpy.ndarray`): Get the polyhedron's edges.
+
+        Results returned as vertex index pairs.
+        """
+        edges = []
+        for face in self.faces:
+            [
+                edges.append((i, j))
```
vyasr, line 374: This is an odd construct. You're using a list comprehension to mimic a for loop, but you're populating it with a bunch of None. Instead, you can do a nested list comprehension where you construct edges in the comprehension rather than appending to it, something like `return [(i, j) for ... for face in self.faces`.

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,36 @@ def vertices(self):
...
+                for i, j in zip(np.append(face, face[0]), np.append(face[1:], face[0]))
+                if (j, i) not in edges
+            ]
+        return edges
+
+    @property
+    def get_edge_vectors(self):
```
vyasr, line 381: We don't name properties with `get_*` prefixes, only functions. Properties should "feel like" accessing an attribute of an object.
```suggestion
    def edge_vectors(self):
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,36 @@ def vertices(self):
...
+        """list(:class:`numpy.ndarray`): Get the polyhedron's edges as vectors."""
+        vecs = []
+        vertices = self.vertices
+        for edge in self.edges:
+            vecs.append(vertices[edge[1]] - vertices[edge[0]])
+        return vecs
```
vyasr, line 387: With numpy indexing this can be simplified a bit (and made faster):
```suggestion
        edges = np.asarray(self.edges)
        return list(self.vertices[edges[:, 1]] - self.vertices[edges[:, 0]])
```

tests/test_polyhedron.py
```diff
@@ -225,6 +225,24 @@ def test_dihedrals():
...
+        "Cube": 1,
+        "Octahedron": np.power(2, 5 / 6) * np.cbrt(3 / 8),
+        "Dodecahedron": np.power(2, 2 / 3) * np.cbrt(1 / (15 + np.sqrt(245))),
+        "Icosahedron": np.cbrt(9 / 5 - 3 / 5 * np.sqrt(5)),
+    }
+    for name, edgelength in known_shapes.items():
+        poly = PlatonicFamily.get_shape(name)
+        if name == "Dodecahedron":
+            poly.merge_faces(rtol=1)
+        for edge in poly.get_edge_vectors:
```
vyasr, line 241: Can we test `edges` as well as the vectors? At the moment that isn't tested at all.

vyasr (changes requested):
Nice work! I have some suggestions for improvement here.

coxeter/shapes/polyhedron.py
janbridley: imageOption 2 


Based on your above feedback, I have come up with these options. The first option is an improved version of the append structure in the previous commit, which implements your suggestion of a nested list comprehension. It takes approximately 86us per iteration, as tested with timeit and 10 million iterations.

The second uses two list comprehensions, one to generate the edges and their reverses, and one to create the final list without reverse elements. It takes approximately 96 microseconds per iteration, tested in the same manner as above.


Note that both methods were formatted with black, so the images above are the formatted structure of the code as written.


Although I tried to come up with a cleaner algorithm for this, I could not make one that (1) returns the edges in an order corresponding to the input faces and (2) was faster than the appending method. I would be happy to hear your input on this if you have any, and if you like either of the above methods I will commit them!

coxeter/shapes/polyhedron.py
janbridley: This is correct, polyhedron faces are ordered wrt/ an outward normal, meaning only the reverse will be added.

coxeter/shapes/polyhedron.py
janbridley: Looks good - I have taken this into account with the changes in the next subconversation.

janbridley:
Looks like CircleCI is failing because of inaccuracies in the dodecahedron's stored data. I feel that we should wait for #177 to be merged before merging this branch (#171), so that the small fixes currently in test_edges can be reverted before this branch goes live. This will result in a cleaner merge.

coxeter/shapes/polyhedron.py
bdice: Let's avoid using a list comprehension if the result is thrown out -- all those `edges.append` calls are returning a list of `[None, None, None, ...]` which is unused. Just use a `for` loop instead of a comprehension in this case, because this logic depends on the current value of the list (`if (j, i) not in edges`). It's a bit faster (no allocation for the throwaway result) and more readable. 

Secondly, we can make this faster if we avoid `np.roll` and "roll our own." 😉 For educational purposes, I wrote up a microbenchmark linked below. Try it out locally! For me, "option 3" is around 3x faster than the other implementations. That said, I would not choose the final implementation based solely on speed. Readability is important.

https://github.com/bdice/python_microbenchmarks/blob/35399c161d2cb35a629e9ceced0f8c7e08abde08/microbenchmarks/faces_to_edges.py

(Another small request -- please write formatted snippets with ` ```python ` [code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) instead of screenshots. It makes it easier to read and propose alternatives.)

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,33 @@ def vertices(self):
...
+
+        Results returned as vertex index lists.
+        """
         return self._faces
 
+    @property
+    def edges(self):
+        """list(:class:`numpy.ndarray`): Get the polyhedron's edges.
+
+        Results returned as vertex index pairs.
```
bdice, line 369: The docs should mention that edges are de-duplicated and that pairs are unordered. However, I'm not really sure if that's what we want. I would think that users would prefer to have control of the result, and de-duplicate and/or determine symmetry `(i, j) == (j, i)` for themselves. Should edges just be `return [(i, j) for f in self.faces for i, j in zip(f, np.roll(f, -1))]`?

bdice: I can see it either way. Both properties would be useful for different things. If you wanted to draw wireframes or make 3D models, you might need the full set of directed face edges (including "duplicates" belonging to different faces).

tests/test_polyhedron.py
```diff
@@ -225,6 +225,29 @@ def test_dihedrals():
...
+        "Cube": 1,
+        "Octahedron": np.power(2, 5 / 6) * np.cbrt(3 / 8),
+        "Dodecahedron": np.power(2, 2 / 3) * np.cbrt(1 / (15 + np.sqrt(245))),
+        "Icosahedron": np.cbrt(9 / 5 - 3 / 5 * np.sqrt(5)),
+    }
+    for name, edgelength in known_shapes.items():
+        poly = PlatonicFamily.get_shape(name)
+        # Test edge_vectors property
+        for edge in poly.edge_vectors:
+            # rtol must be lowered to accomodate small inaccuracies in vertex locations
```
bdice, line 241:
```suggestion
            # rtol must be lowered to accommodate small inaccuracies in vertex locations
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,33 @@ def vertices(self):
...
+                for i, j in zip(np.append(face, face[0]), np.append(face[1:], face[0]))
+                if (j, i) not in edges
+            ]
+        return edges
+
+    @property
+    def edge_vectors(self):
+        """list(:class:`numpy.ndarray`): Get the polyhedron's edges as vectors."""
+        edges = np.asarray(self.edges)
+        return list(self.vertices[edges[:, 1]] - self.vertices[edges[:, 0]])
```
bdice, line 384: Why return this as a `list`? Ideally we can keep everything as NumPy arrays unless it's a ragged array (e.g. faces may have different numbers of vertices in them).

Also, because `self.vertices` is a NumPy array, you don't have to convert `edges`. You can index it with a list and it'll do fancy NumPy indexing.

```suggestion
        edges = self.edges
        return self.vertices[edges[:, 1]] - self.vertices[edges[:, 0]]
```

tests/test_polyhedron.py
janbridley: Because the rtol change was removed, I have deleted this line entirely - thank you for catching this!

coxeter/shapes/polyhedron.py
janbridley: My personal feeling is that the 'cleanest' output should be accessible directly from the package, and the (slightly) more complicated case can be done by the user. That being said, I agree with your sentiment that neither method is incorrect. @vyasr I would appreciate your input here as well

bdice: @janbridley Can you also give context on your motivation for this PR? What does your use case require?

janbridley: Also note that edges generated here are ordered: edges move clockwise around the surface of a face, and faces are ordered counterclockwise with respect to an outward normal. Note that using j,i pairs reverses the direction of each vector, but keeps their clockwise ordering.

janbridley: I see - thank you for clarifying. I was hoping to come up with a logical way to order the edges, but as you mention there isn't really a strategy that is applicable to all convex polyhedra. The [Eulerian path](https://en.wikipedia.org/wiki/Eulerian_path) is probably closest to the result I was hoping for, but cannot be calculated for a general convex polyhedra. Based on this, how do we feel about returning the edges of the polyhedra as a set? I think that could help convey the lack of explicit ordering for the property. 

In addition, I think that including a second property that returns edges and their (j,i) duplicates could both alleviate confusion and provide easy access to both sets of data - what are your thoughts on that?

Tobias-Dwyer: I like Jen's solution above, at the very least we should have the option to have both edge types and in the past it's been annoying to get multiple copies of an edge. Having both (or at least having the duplicate edges as a method) would immediately clue users in that edges are not as simple as they might first expect.

bdice: Sets of sets would be the most Python-friendly way to indicate this that I can think of. However, you'll run into a problem: sets are mutable, and thus not hashable. You'd have to have a set of `frozenset` objects (immutable sets, which are hashable). We could return a NumPy array, sorted and with only pairs where `i < j`.

bdice: I'd advise against multiple properties unless we have a really clear use case. I'm still not sure if I fully understand what the intended use case is for this property. How you do want to use the `edges` property?

janbridley: ``Edges`` allows for the calculation of each component of the asphericity of a polyhedra, allowing for a determination of the relative contribution of each feature. ``Edge_vectors`` allows for the calculation of angles and distances between specific edges on a polyhedron, which can be useful for simulations of highly anisotropic particles. @Tobias-Dwyer if you have any contributions I'd love to hear them!

coxeter/shapes/polyhedron.py
vyasr: Yup you're right I didn't account for the deduplication process needing to access the previously existing ones. A simple for loop is probably easiest alternative. I would definitely go for higher readability over reimplementing something like roll. coxeter is pretty suboptimal performance-wise anyway.

coxeter/shapes/polyhedron.py
vyasr: > My personal feeling is that the 'cleanest' output should be accessible directly from the package, and the (slightly) more complicated case can be done by the user.

This is the right way to think about it. The package shouldn't be overloaded with a dozen variations on essentially the same property. From that perspective, given that as you've pointed out there are multiple possibilities for exactly what is required in different situations, would we be better served by including more documentation of how to compute edges in the documentation of the `faces` property? Otherwise we could easily proliferate multiple options based on considerations of ordering and uniqueness.

coxeter/shapes/polyhedron.py
janbridley: Looks good - The roll1 function is impressive but I agree it is probably not the cleanest option.


> (Another small request -- please write formatted snippets with `` ```python `` [code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) instead of screenshots. It makes it easier to read and propose alternatives.)

Noted - thank you for the information. The following code has been formatted as requested.

**Option 1: most readable**

```python
def edges():
    return {(i, j) for face in self.faces for i, j in zip(face, np.roll(face, -1)) if i < j}
```

**Option 2: fastest (by about an order of magnitude)**
```python
def edges():
  return{
      (face[i], face[(i + 1) % len(face)])
      for face in self.faces
      for i in range(len(face))
      if face[i] < face[(i + 1) % len(face)]
  }
```

As above, I think the more readable option 1 is the better choice for this package.

coxeter/shapes/polyhedron.py
janbridley: Overall, I feel that some variant of ``edges`` should be included with the package - along with vertices and faces, edges are a defining [element](https://en.wikipedia.org/wiki/Polytope#Elements) of polyhedra, and the number of either _(i,j)_ or _(j,i)_ edges allows for the correct calculation of the [Euler characteristic](https://en.wikipedia.org/wiki/Euler_characteristic), an intrinsic property of polyhedra in general. I agree that cluttering the package is undesirable, but I feel that including documentation on how to calculate edges is no cleaner than just efficiently implementing that method. 

Looking at other widely used geometry packages, it seems like selecting edges with _(i<j)_ is probably the best way forward for Coxeter. Mathematica's ``PolyhedronData[]`` is ordered this way, and it is probably the least confusing approach. [CGAL](https://doc.cgal.org/latest/Polyhedron/classCGAL_1_1Polyhedron__3_1_1Halfedge.html) provides the ``Halfedge``trait for polyhedra, which is equivalent to our _(i,j)_ or _(j,i)_ pairs, and allow users to select the direction and ordering of edges. However, CGAL's ``Polyhedron_3`` does not calculate edges for an entire polyhedra at once: rather, there are independent member functions to find local half-edges from a vertex, edge, or face on the surface. This approach is probably too verbose for Coxeter, but I am not opposed to renaming our method ``halfedges`` if that would avoid confusion. 

Based on the above, I think the best approach is probably to return a ``set`` of edges with _(i<j)_ and each edge included once. The documentation will be updated to better describe the method, and I have included two options for the new code [above](https://github.com/glotzerlab/coxeter/pull/171#discussion_r1126805403).

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,40 @@ def vertices(self):
...
+
+    @property
+    def edge_vectors(self):
+        """set(tuple(float,float,float)): Get the polyhedron's edges as vectors."""
+        return [
+            np.subtract(
+                *self.vertices[[j, i]],
+            )
+            for (i, j) in self.edges
+        ]
```
vyasr, line 391: While a set makes sense for returning the edges, it doesn't really make sense for the vectors IMO. The edge list is already deduplicated, and vectors are floats so a set doesn't typically help much anyway without rounding the values. Rather than a set of arrays, I would recommend a simple 2D numpy array. The result will also be faster.
```
In [11]: %timeit cube.edge_vectors
116 µs ± 1.92 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [12]: %%timeit
    ...: edges = np.asarray(list(cube.edges))
    ...: cube.vertices[edges[:, 1]] - cube.vertices[edges[:, 0]]
    ...:
    ...:
69.4 µs ± 1.6 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +356,40 @@ def vertices(self):
...
+        Results returned as vertex index lists.
+        """
         return self._faces
 
+    @property
+    def edges(self):
+        """set(tuple(int,int)): Get the polyhedron's edges.
+
+        Results returned as vertex index pairs,  with each edge of the polyhedron
+        included exactly once.  Edge (i,j) pairs are ordered by vertex index with i<j.
```
vyasr, line 370: Sets (unlike dicts) are not insertion ordered. While you may observe that in some cases depending on your Python installation, there is no guarantee (see e.g. [this StackOverflow post](https://stackoverflow.com/questions/61414947/why-dont-python-sets-preserve-insertion-order)). If we want to promise ordering (and I believe that we should), we should probably wrap the returned set in a `sorted` call so that the output is a list that is consistently ordered. This is especially true since if we are providing both `edges` and `edge_vectors` it would be confusing to users if `edge_vectors[i]` is not guaranteed to map to `edges[i]`.

vyasr (changes requested):

coxeter/shapes/polyhedron.py
janbridley: ``sorted()`` should convert the set to a list, making it subscriptable and ordering it such that ``edges`` and ``edge_vectors`` map as expected. Based on this and your message [above](https://github.com/glotzerlab/coxeter/pull/171#discussion_r1149424714), are you suggesting that both ``edges`` and ``edge_vectors`` return a Numpy array of _(i,j)_ pairs sorted by increasing _i_? This seems reasonable and is in line with Mathematica's ``PolyhedronData[]``, so I can certainly make those changes

vyasr: Correct, `sorted` will make it a list. I would be fine with either returning a list or converting that list to a numpy array, with a weak pref for the latter.

DomFijan:
@vyasr @bdice 
From reading the discussions it seems that the final solution would be to return a numpy array of sorted pairs? After that docs should be updated accordingly? Is this correct?

bdice:
@DomFijan Yeah, I think there were a few proposals above. One was to return a `set` of `tuple` of `int`. Another was to return a NumPy array of the same data. I think I lean towards a NumPy array, and just document that the data is in sorted order with `i < j`.

What I mean concretely is this:

```python
from functools import cached_property  # Requires Python 3.8 or newer, which is fine

class Polyhedron:
    @cached_property
    def edges(self):
        """:class:`numpy.ndarray`: Get the polyhedron's edges.

        Results returned as vertex index pairs,  with each edge of the polyhedron
        included exactly once.  Edge (i,j) pairs are ordered by vertex index with i<j.
        """
        all_edge_pairs = np.array([[i, j] for face in self.faces for i, j in zip(face, np.roll(face, -1))])
        all_edge_pairs.sort(axis=1)
        ij_pairs = np.unique(all_edge_pairs, axis=0)
        # Make the edge data read-only so that the cached property of this instance cannot be edited
        ij_pairs.flags.writeable = False
        return ij_pairs
```

I haven't tested this explicitly but I suspect you can tidy that up with pre-commit and it should be good to go. The best part about using a NumPy array as the return value is that you can use it to index the vertices. (You can't do this as easily if you return a set of tuples of integers.) Then you can compute the edge vectors with one line:

```python
edge_vectors = vertices[edges[:, 1]] - vertices[edges[:, 0]]
```

vyasr:
I advocated against a set because we want to preserve order, so the options would be a numpy array or a list. Between those two I would lean towards a numpy array, yeah.

janbridley:
> @DomFijan Yeah, I think there were a few proposals above. One was to return a `set` of `tuple` of `int`. Another was to return a NumPy array of the same data. I think I lean towards a NumPy array, and just document that the data is in sorted order with `i < j`.
> 
> What I mean concretely is this:
> 
> ```python
> from functools import cached_property  # Requires Python 3.8 or newer, which is fine
> 
> class Polyhedron:
>     @cached_property
>     def edges(self):
>         """:class:`numpy.ndarray`: Get the polyhedron's edges.
> 
>         Results returned as vertex index pairs,  with each edge of the polyhedron
>         included exactly once.  Edge (i,j) pairs are ordered by vertex index with i<j.
>         """
>         all_edge_pairs = np.array([[i, j] for face in self.faces for i, j in zip(face, np.roll(face, -1))])
>         all_edge_pairs.sort(axis=1)
>         ij_pairs = np.unique(all_edge_pairs, axis=0)
>         # Make the edge data read-only so that the cached property of this instance cannot be edited
>         ij_pairs.flags.writeable = False
>         return ij_pairs
> ```
> 
> I haven't tested this explicitly but I suspect you can tidy that up with pre-commit and it should be good to go. The best part about using a NumPy array as the return value is that you can use it to index the vertices. (You can't do this as easily if you return a set of tuples of integers.) Then you can compute the edge vectors with one line:
> 
> ```python
> edge_vectors = vertices[edges[:, 1]] - vertices[edges[:, 0]]
> ```

This is clean code, but is ~50% slower than the current method. However, the sort that np.unique performs does give you the right answer. The following code adapts this to improve the method's performance while also cleaning up my previous code:

``python
> from functools import cached_property  # Requires Python 3.8 or newer, which is fine
> 
> class Polyhedron:
>     @cached_property
>     def edges(self):
>         """:class:`numpy.ndarray`: Get the polyhedron's edges.
> 
>         Results returned as vertex index pairs,  with each edge of the polyhedron
>         included exactly once.  Edge (i,j) pairs are ordered by vertex index with i<j.
>         """
>         ij_pairs = np.array([[i, j] for face in self.faces for i, j in zip(face, np.roll(face, -1)) if i < j])  
>         sorted_indices = np.lexsort(ij_pairs.T[::-1])
>         sorted_ij_pairs = ij_pairs[sorted_indices]    
>         sorted_ij_pairs.flags.writeable = False
>         return sorted_ij_pairs
  

```

And, as you mentioned - the edge_vectors code has been

File: coxeter/shapes/polyhedron.py [hunk 12]
```diff
@@ -356,9 +357,45 @@ def vertices(self):
...
+
+    @property
+    def edge_vectors(self):
+        """list(tuple(float,float,float)): Get the polyhedron's edges as vectors."""
+        return self.vertices[self.edges[:, 1]] - self.vertices[self.edges[:, 0]]
+
+    @property
+    def num_edges(self):
+        """int: Get the number of edges."""
+        return len(self.edges)
```
bdice, line 397: This can be calculated without the need to compute `self.edges` at all. `F + V = E + 2`, per Euler's formula. 😄 I'd suggest:
```suggestion
        return self.num_vertices + self.num_faces - 2
```
This avoids doing the face iterations for complex shapes.

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +357,45 @@ def vertices(self):
...
-        """list(:class:`numpy.ndarray`): Get the polyhedron's faces."""
+        """list(:class:`numpy.ndarray`): Get the polyhedron's faces.
+
+        Results returned as vertex index lists.
+        """
         return self._faces
 
+    @cached_property
+    def edges(self):
+        """set(tuple(int,int)): Get the polyhedron's edges.
```
bdice, line 368:
```suggestion
        """:class:`numpy.ndarray`: Get the polyhedron's edges.
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +357,45 @@ def vertices(self):
...
+        sorted_ij_pairs = ij_pairs[sorted_indices]
+        # Make edge data read-only so that the cached property of this instance
+        # cannot be edited
+        sorted_ij_pairs.flags.writeable = False
+
+        return sorted_ij_pairs
+
+    @property
+    def edge_vectors(self):
+        """list(tuple(float,float,float)): Get the polyhedron's edges as vectors."""
```
bdice, line 391:
```suggestion
        """:class:`numpy.ndarray`: Get the polyhedron's edges as vectors."""
```

tests/test_polyhedron.py
```diff
@@ -339,6 +339,44 @@ def test___repr__():
...
+        poly = PlatonicFamily.get_shape(name)
+        # Test that the correct number of edges has been found
+        assert number_of_edges[name] == poly.num_edges
+        assert number_of_edges[name] == len(poly.edge_vectors)
+
+        # Test edge_vectors property
+        for edge in poly.edge_vectors:
+            assert np.isclose(np.linalg.norm(edge), edgelength)
+
+        # Test edges property
```
bdice, line 369: To test the `edges` property, I would consider some of its invariants / properties:
- assert that the first column of `i` indices is sorted
- assert that the the indices are ordered `i < j`
- assert that the array has an integer dtype
- etc.

janbridley: See [986fa8d](https://github.com/glotzerlab/coxeter/pull/171/commits/986fa8d2132fb99d43e47bdf2ad08c59e2620078)

coxeter/shapes/polyhedron.py
janbridley: This is only strictly true for convex shapes - I have added a method to convex_polyhedron that makes use of the Euler characteristic, and have added additional pytests (see [d3f5203](https://github.com/glotzerlab/coxeter/pull/171/commits/d3f520379977f239033fb7351039d9336884b3ef))

DomFijan:
@bdice, any chance you can take a look at this before the end of Wednesday next week if you have the time? If not, I can jump in so we can finish this PR. There are a couple more tasks that depend on this that we want to start moving along. :)

Credits.rst
```diff
@@ -112,6 +112,7 @@ Jen Bradley
 * Added shape families for Archimedean, Catalan, and Johnson solids.
 * Added shape family for prisms and antiprisms.
 * Added shape family for equilateral pyramids and dipyramids.
+* Added edges, edge_vectors, and num_edges methods
```
bdice, line 115:
```suggestion
* Added edges, edge_vectors, and num_edges methods.
```

coxeter/shapes/polyhedron.py
```diff
@@ -356,9 +357,45 @@ def vertices(self):
...
+        """:class:`numpy.ndarray`: Get the polyhedron's edges.
+
+        Results returned as vertex index pairs,  with each edge of the polyhedron
+        included exactly once.  Edge (i,j) pairs are ordered by vertex index with i<j.
+        """
+        ij_pairs = np.array(
+            [
+                [i, j]
+                for face in self.faces
+                for i, j in zip(face, np.roll(face, -1))
```
bdice, line 377: Just making sure — the face indices are ordered consistently, right? In other words, this code relies on face indices being sorted in such a way that any shared edge is in the order (i, j) for one face and (j, i) for the other face sharing that edge. Is that a safe assumption from the rest of the code (it has been a while since I’ve read it). If this isn’t handled properly by the construction of the shape, then you’ll get missing edges. Are there any degenerate cases where this doesn’t happen automatically (can a polyhedron have one face in coxeter)?

coxeter/shapes/polyhedron.py
bdice: Good point. Thanks for catching that. I’ll resolve this.

tests/test_polyhedron.py
```diff
@@ -339,6 +339,96 @@ def test___repr__():
...
+    named_johnson_mark,
+    named_prismantiprism_mark,
+    named_pyramiddipyramid_mark,
+)
+def test_edges(poly):
+    # Check that the first column is in ascending order.
+    assert np.all(np.diff(poly.edges[:, 0]) >= 0)
+
+    # Check that all items in the first column are greater than those in the second.
+    assert np.all(np.diff(poly.edges, axis=1) >= 0)
```
bdice, line 355: i-i edges should be illegal so this can be a strict inequality.
```suggestion
    assert np.all(np.diff(poly.edges, axis=1) > 0)
```

tests/test_polyhedron.py
```diff
@@ -339,6 +339,96 @@ def test___repr__():
...
+        [
+            np.all(np.diff(edges[edges[:, 0] == value, 1]) >= 0)
+            for value in unique_values
+        ]
+    )
+
+    # Check that there are no duplicate edges. This also double-checks the sorting
+    assert np.all(np.unique(poly.edges, axis=1) == poly.edges)
+
+    # Check that the edges are immutable
```
bdice, line 371: This can be simpler if you want. I would just use pytest’s “assert raises” context manager, wrapping the assignment.

bdice (approved):
A few small suggestions / verifications but otherwise LGTM. Feel free to merge once you are happy with it.

coxeter/shapes/polyhedron.py
janbridley: I am unsure exactly what the limits of Polyhedron are: for example, if self-intersecting shapes, shapes with holes, etc work properly. That being said, faces are ordered and edges are built using another method during initialization so it must work for the vast majority of use cases! I can take a deeper look in the next few days.

janbridley: Conplex polyhedra (within limits) seem to work - self intersecting polyhedra, polyhedra with holes, and simple non convex (e.g. stellated) polyhedra all compute edges as expected. However, there are cases where volumes, moments of inertia, and other properties are incorrect. There are some cases where edges does NOT work - but in these cases,  many other methods break down. Polygons (one-face polyhedron), dihedra, and nonclosed solids tend to break this method but they also return null values for any method that includes volume. Some also have erroneous behavior in the centroid or inertia tensor calculation. I will create an issue for some of these edge cases in general, but overall the new edge methods are at least as robust as the other core methods.

bdice: Good. It’s good to know the limitations — and which ones you are willing to treat as undefined behavior vs. raising an error or “fix” in some way.

DomFijan:
@bdice thanks for the prompt reaction :)

PR #178, opened 2023-02-13
Fix minimal bounding sphere

janbridley:
This is a test PR to check a bug in CircleCI. No action is required


coxeter/shapes/polyhedron.py
```diff
@@ -550,27 +550,40 @@ def minimal_bounding_sphere(self):
...
+
+        # The following subfunction checks if all vertices lie within the
+        # minimal bounding sphere, to a tolerance of 1e-15.
+        def verify_inside(points, center, radius, eps=1e-15):
+            return np.all(np.linalg.norm(points - center, axis=1) <= radius + eps)
+
+        # Simple polyhedra (e.g. dodecahedra) will require far less than 200 attempts
+        # to compute a correct miniball. Polyhedra with augmentations and large numbers
+        # of vertices will take more, but none are likely to exceed 200 attempts. Worst-
+        # case runtime is approximately 0.25s, and occurrs in ~5/1e5 samples.
```
bdice, line 570:
```suggestion
        # case runtime is approximately 0.25s, and occurs in ~5/1e5 samples.
```

coxeter/shapes/polyhedron.py
```diff
@@ -550,27 +550,40 @@ def minimal_bounding_sphere(self):
...
-                center, r2 = miniball.get_bounding_ball(vertices)
+                center, r2 = cyminiball.compute(vertices)
+                assert verify_inside(vertices, center, np.sqrt(r2))
                 break
             except np.linalg.LinAlgError:
                 current_rotation = rowan.random.rand(1)
                 vertices = rowan.rotate(current_rotation, vertices)
+            except AssertionError:
+                current_rotation = rowan.random.rand(1)
+                vertices = rowan.rotate(current_rotation, vertices)
```
bdice, line 586: To reduce numerical error, let’s apply the new random rotation to the original vertices instead of rotating the same vertices multiple times.
```suggestion
                vertices = rowan.rotate(current_rotation, self.vertices)
```

coxeter/shapes/polyhedron.py
```diff
@@ -550,27 +550,40 @@ def minimal_bounding_sphere(self):
...
-        # random rotation will usually fix the issue.
-        max_attempts = 10
+        # The miniball algorithm (in this case, Gärtner's miniball) can
+        # experience numerical instability of ~10x machine epsilon. If an incorrect
+        # miniball is found, applying random rotations will usually fix the issue.
+
+        # The following subfunction checks if all vertices lie within the
+        # minimal bounding sphere, to a tolerance of 1e-15.
+        def verify_inside(points, center, radius, eps=1e-15):
+            return np.all(np.linalg.norm(points - center, axis=1) <= radius + eps)
```
bdice, line 565: Should we compare squared distances? If we avoid the square root in the `norm` here and just use sum-of-squares, we can also reuse the value of `r2` below without incurring another sqrt. Just depends on whether you think the epsilon is more meaningful in terms of distance rather than distance squared, I suppose.

coxeter/shapes/polyhedron.py
```diff
@@ -550,27 +550,40 @@ def minimal_bounding_sphere(self):
...
             attempt += 1
             try:
-                center, r2 = miniball.get_bounding_ball(vertices)
+                center, r2 = cyminiball.compute(vertices)
+                assert verify_inside(vertices, center, np.sqrt(r2))
                 break
             except np.linalg.LinAlgError:
                 current_rotation = rowan.random.rand(1)
                 vertices = rowan.rotate(current_rotation, vertices)
+            except AssertionError:
```
bdice, line 584: Let’s combine this with the previous LinAlgError and handle both at the same time. Should look something like:

```suggestion
            except (np.linalg.LinAlgError, AssertionError):
```

bdice (changes requested):
Nice work. The issue and PR descriptions were very helpful. I have some comments to discuss.

bdice:
@janbridley FYI, I merged in the upstream to fix CI failures.

joaander (changes requested):
_cyminibal_ is GPL licensed. See details in #179.

coxeter/shapes/polyhedron.py
janbridley: This is excellent, thank you

coxeter/shapes/polyhedron.py
janbridley: For small distances (e.g. for points that are very close to the bounding sphere), ε will be more significant and therefore represent a larger tolerance range for accepted points. This is magnified for square distances, so if we do make this change we should probably test a smaller ε. I will set this aside while I figure out _hbf/miniball_ and we can revisit later

PR #186, opened 2023-06-26
Coxeter shapes can be exported as a JSON serializable object

janbridley:
To improve compatibility of Coxeter with HOOMD-Blue and the Signac framework, a to_hoomd() export option could be added. This is a similar idea to the gsd_shape_spec, but contains significantly more data to aid in the writing and automation of simulation scripts (namely, the volume, inertia tensor, and centroid). Although this object can be used in signac statepoints, it is HIGHLY recommended to save the data to the signac project document instead.

```python
import signac
from coxeter.families import PlatonicFamily

cube = PlatonicFamily.get_shape("Cube")


project = signac.init_project()

statepoint = dict(N=64, seed=1, vf=0.60, types=["cube"])

job = project.open_job(statepoint)
job.document.poly = cube.to_hoomd()
```

Once loaded into the statepoint, many shape attributes are accessible from the dict:
```python
job.document.poly.vertices
>>>[[-0.5, -0.5, -0.5],
    [-0.5, -0.5,  0.5],
    [-0.5,  0.5, -0.5],
    [-0.5,  0.5,  0.5],
    [ 0.5, -0.5, -0.5],
    [ 0.5, -0.5,  0.5],
    [ 0.5,  0.5, -0.5],
    [ 0.5,  0.5,  0.5]]

job.document.poly.centroid
>>>[0.0, 0.0, 0.0]

job.document.poly.gsd_shape_spec
>>>{'type': 'ConvexPolyhedron', 'vertices': [[-0.5, -0.5, -0.5], [-0.5, -0.5, 0.5], [-0.5, 0.5, -0.5], [-0.5, 0.5, 0.5], [0.5, -0.5, -0.5], [0.5, -0.5, 0.5], [0.5, 0.5, -0.5], [0.5, 0.5, 0.5]]}

```

All coxeter.shapes store their defining attributes (for example, radius for a Sphere and vertices for a ConvexPolyhedron). All shapes also store their `centroid` in the new dict, as well as a few other useful properties.

## Description
* Added to_hoomd() export function
* Added pytests

## Motivation and Context
Signac scripts using this new structure can better interface with molecular dynamics software, especially HOOMD. For example, [gsd-formatted](https://hoomd-blue.readthedocs.io/en/latest/tutorial/02-Logging/03-Storing-Particle-Shape.html) `type_shapes` can be added to a logger directly from the job document via `job.document.poly.gsd_type_shapes`. Similarly, `hoomd.hpmc` integrators can be initialized from data stored within the `poly` object and a shape's `inertia_tensor` can be directly added to `particles.moment_inertia`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

<sup>1</sup>The change has the potential to break existing functionality. Users who rely on the default-generated __dict__ method may find keys have different names after this change. 

## Checklist:


coxeter/shapes/base_classes.py
```diff
@@ -193,6 +194,41 @@ def to_plato_scene(self, backend="matplotlib", scene=None, scene_kwargs=None):
             scene.add_primitive(prim)
         return scene
 
+    def _data(self):
+        return self.__dict__
+
+    def __dict__(self):
```
bdice, line 200: Don’t use `__dict__` for this. I would recommend using another attribute/method name that is not reserved by the Python data model. Something like `to_dict(self)` or `as_dict(self)` or `serialize(self)`. The `object.__dict__` attribute is considered special and is reserved as a read-only attribute. https://docs.python.org/3/library/stdtypes.html#special-attributes

Additionally, be aware that serialization of shapes to JSON is lossy (float to string conversion).

coxeter/shapes/base_classes.py
```diff
@@ -193,6 +194,41 @@ def to_plato_scene(self, backend="matplotlib", scene=None, scene_kwargs=None):
...
+
+        Returns:
+           dict_items: List of items stored in self.__dict__.
+
+        :meta private:
+        """
+        return self.__dict__.items()
+
+
+Mapping.register(Shape)
```
bdice, line 230: Is registration necessary here? I would prefer to see the class meet the ABC Mapping interface requirements to satisfy this implicitly, rather than explicitly stating “this is a Mapping” by fiat.

coxeter/shapes/circle.py
```diff
@@ -228,3 +228,17 @@ def _plato_primitive(self, backend):
...
+    @property
+    def _data(self):
+        return self.__dict__
+
+    @property
+    def __dict__(self):
+        return {
+            "radius": self.radius,
+            "centroid": self.centroid.tolist(),
+            "inertia_tensor": self.inertia_tensor,
```
bdice, line 241: I would prefer to see shape serializations contain only the minimal amount of information needed to reconstruct the shape as an object. Coxeter should be used to compute an inertia tensor when it is needed, from the inputs to the class constructor.

bdice (changes requested):
This deviates from the Python object model in a few key ways. I would hesitate to merge this without a significant design change. Perhaps discussing the intended use cases would help point towards a solution that aligns better with the Python object model while still being useful for serializing shapes for use in signac data spaces.

bdice:
The GSD shape specification is almost exactly what I would recommend as a serialization format. Can you use the `gsd_shape_spec` attribute to serialize the shape as a dict, save that in the statepoint, and use `coxeter.from_gsd_type_shapes` to deserialize the dict into a shape for further computation? If there are shapes this doesn’t handle, we might be able to expand the GSD shape spec, but last I knew this should be sufficient.

coxeter/shapes/base_classes.py
janbridley: Sounds good - thank you for the feedback on this, and I will look into a more specific (and nonreserved) method

coxeter/shapes/circle.py
janbridley: See [b1e5165](https://github.com/glotzerlab/coxeter/pull/186/commits/b1e51659a95cec39f696e1c4a3861883e2f03300)

janbridley:
This PR requires some thought into the best manner to approach this problem, and I do not believe it is particularly high priority right now. I have marked it as a draft to direct focus toward more important tasks.

janbridley:
I will reopen this issue as a new PR with significant changes to the internal implementation

PR #187, opened 2023-08-03
Speed up point in polygon with numpy

DomFijan:
## Description
I reimplemented winding number calculator using numpy.
This is technically a breaking change since points on border are now treated as outside, in line with the methodology. To have the same type of functionality as before additional conditions would have to be checked reducing the speedup. It might be prudent to make special functions which check weather points are on edge and/or on vertex.

## Motivation and Context
It will significantly speed up the point in polygon check, at the cost of memory. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:


DomFijan:
Few things I would specifically like feedback on:

- I have just dumped the docstring from the old file into the is_inside function. I'm looking for feedback on this. Should it be cut down?
- I'm also looking for feedback on the API in terms of the breaking change. Should I add another function that checks if points are on the boundary to supplement this? 
I have played with the boundary check as it was implemented before and it will only work if a point is exactly on the boundary in line with machine precision. A more useful check (which I often use) is to have a check with some tolerance. 
Would changes re this be better suited for another PR?
- Another thing I'm unsure about is if the forcing of vertex points onto a plane is dragging down the performance of the function as well. I guess this is a problem only in 3D. Maybe a better thing would be to first check if the polygon is embedded in 2D first and if not then align and project?

janbridley:
Looks good, I am reviewing now. 

> * I have just dumped the docstring from the old file into the is_inside function. I'm looking for feedback on this. Should it be cut down?

I am going to reformat the docstring to properly use sphinx-math/latex formatting, and add that as a comment. I think that the text is a little too long, but we should definitely keep a lot of the detail you've included because this isn't a simple method. Once I have the math formatted, we can review for more concise wording?

> * I'm also looking for feedback on the API in terms of the breaking change. Should I add another function that checks if points are on the boundary to supplement this?
>   I have played with the boundary check as it was implemented before and it will only work if a point is exactly on the boundary in line with machine precision. A more useful check (which I often use) is to have a check with some tolerance.
>   Would changes re this be better suited for another PR?
I think that better specifying the tolerances for points near the boundary is a good addition to this PR, because it will (a) help address numerical imprecision issues that are currently mentioned in the docstring and (b) provide more clarity on the actual behavior near the boundary, which is a big portion of the breaking change.

> * Another thing I'm unsure about is if the forcing of vertex points onto a plane is dragging down the performance of the function as well. I guess this is a problem only in 3D. Maybe a better thing would be to first check if the polygon is embedded in 2D first and if not then align and project?
This would be a good place for continued work on #149, and I think is something that could be improved later?

coxeter/shapes/polygon.py
```diff
@@ -667,12 +666,50 @@ def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
     def is_inside(self, points):
```
bdice, line 666: We could consider adding a tolerance, but maybe better for a separate PR.

bdice (commented):
Since this PR is primarily about optimization, can you share some benchmarks?

DomFijan:
> Since this PR is primarily about optimization, can you share some benchmarks?

For uptick on speed I tested the following code:
```
import coxeter
import numpy as np

polygon = coxeter.shapes.Polygon([[0.4532, 1.653], [-1.626, 0.4532], [-1.626, -1.626], [0.4532, -1.626], [1.653, 0.4532]])

# make 1000 random points inside or outside of polygon
points = np.random.uniform(-2, 2, size=(1000, 2))
# add a column of zeros to the points 
points = np.hstack((points, np.zeros((points.shape[0], 1))))
%timeit inside = polygon.is_inside(points)
```
on master branch I get the following: `10.9 ms ± 347 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`
on branch update_point_in_polygon  I get the following: `394 µs ± 38.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

DomFijan:
@janbridley  and I discussed offline that it might be useful to have a winding number calculator function for other use-cases. I agree that this should be done, but might be better to do in a separate PR once point in polyhedron is converted to efficient numpy implementation. 
I took a peak at the point in polyhedron function and it seems that a lot of code used in this PR can be reused for point in polyhedron check. The winding number calculator might need some tweaks to be compatible with both functions. 
I propose that I first convert point in polyhedron to numpy arrays, then abstract out the winding number calculator for use in both functions. Then I would first have a PR that implements the winding number calculator abstraction, followed by a point in polyhedron PR. Any thoughts?

tests/test_polygon.py
```diff
@@ -506,12 +506,81 @@ def test_is_inside(convex_square):
     assert convex_square.is_inside(convex_square.center)
     assert rotated_square.is_inside(rotated_square.center)
 
-    @given(floats(0, 1), floats(0, 1))
-    def testfun(x, y):
-        assert convex_square.is_inside([[x, y, 0]])
-        assert rotated_square.is_inside([[x, y, 0]])
-
-    testfun()
+    point1 = np.array([0.0, 0.0])
```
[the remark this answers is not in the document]
DomFijan: reverted. I had to figure the precision limits for this approach.

coxeter/shapes/polygon.py
DomFijan: addressing @janbridley comment on this as well. The method implemented here will correctly address any point inside the boundary, all the way to the smallest positive normalized floating-point number which should be 2.2250738585072014e-308. Beyond this numerical instabilities are possible according to my testing, but I think this is very reasonable. Winding number can be used to determine if a point is exactly on the boundary within machine precision. This was done previously by catching an exception which I have removed for performance reasons and because in my own experience this is not very useful. I propose that a more useful approach would be to have a separate function that determines if a point is on boundary within some tolerance. The optimal way to categorize points as inside/outside/on boundary/on vertex would be the following:

1. check if points are on a vertex within some tolerance (cheap)
2. for points not on vertex check if they are in an edge (slightly more expensive)
3. for points not on vertex or edge check if inside polygon (most expensive)
 
Each of these should be a separate function and we could make another function that returns a point by point categorization of the above scenario. In all the cases tolerance would only apply for the point on vertex or point on edge checks, since point in polygon is exact (within machine precision). I think this effort should be in a separate PR. I found this workflow works well for my use case. Do you have any thoughts on this (also @janbridley )?

bdice: This sounds good to me. Denormalized floating point values are out of scope for this computation. 😄

It sounds like you're basically computing tolerance as-if the shape is a spheropolygon. That's fine by me. I would explicitly skip steps 1 & 2 (vertex checks and edge checks) when `tolerance <= 0`. All other parts of your plan sound like a great design.

bdice:
For the winding-number and point-in-polyhedron -- your proposal sounds solid to me.

coxeter/shapes/polygon.py
janbridley: I agree that additional functions to test for points on the boundary would be extremely useful. I also think a tiered approach (as you lay out above) is a good way to avoid wasted computational time in a bulk check, and provide granular access to specific checks if desired. Sounds like a good plan!

coxeter/shapes/polygon.py
```diff
@@ -667,12 +666,50 @@ def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
...
         """Determine whether points are contained in this polygon.
```
janbridley, line 667:
```suggest
        """Simple point-in-polygon algorithm based on winding number.

        The code in this function is based on implementation in :cite:`Dickinson2019` 
        which is licensed under the BSD-3 license. 
        
        Given a closed, possibly non-simple polygon described as a list of vertices in 
        :math:`\mathbb{R}^2` and a point that doesn't lie directly on the path of the 
        polygon, we'd like to compute the winding number of the polygon around the 
        point. To achieve this, we place the point at the origin. Divide the remainder 
        of the plane (i.e., :math:`\mathbb{R}^2` minus the origin) into two halves, 
        :math:`L` and :math:`R`, defined as follows:

        .. math::
            L = {(x, y) | x < 0 \lor x = 0 \land y < 0}

            R = {(x, y) | x > 0 \lor x = 0 \land y > 0}

        That is, :math:`R` contains all points with argument in the half-closed
        interval :math:`\\left[-\\frac{\pi}{2},\\frac{\pi}{2}\\right)`, and :math:`L` contains all 
        others.  Note that with these definitions, :math:`L` and :math:`R` are both 
        convex: a line segment between two points in :math:`R` lies entirely 
        in :math:`R`, and similarly for :math:`L`.  In particular, a line segment 
        between two points can only pass through the origin if one of those points is 
        in :math:`L` and the other in :math:`R`. Now, we follow the edges of the 
        polygon, keeping track of how many times we move between :math:`L` 
        and :math:`R`.  For each move from :math:`L` to :math:`R` (or vice versa),
        we also need to compute whether the edge passes *above* or *below* the origin,
        to compute its contribution to the total winding number.  From the comment
        above, we can safely ignore all edges that lie entirely within either :math:`L` 
        or :math:`R`.
```

coxeter/shapes/polygon.py
```diff
@@ -667,12 +666,50 @@ def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
...
 
+        The code in this function is based on implementation in
```
janbridley, line 669:
```suggestion
        """Simple point-in-polygon algorithm based on winding number.

        The code in this function is based on implementation in :cite:`Dickinson2019` 
        which is licensed under the BSD-3 license. 
        
        Given a closed, possibly non-simple polygon described as a list of vertices in 
        :math:`\mathbb{R}^2` and a point that doesn't lie directly on the path of the 
        polygon, we'd like to compute the winding number of the polygon around the 
        point. To achieve this, we place the point at the origin. Divide the remainder 
        of the plane (i.e., :math:`\mathbb{R}^2` minus the origin) into two halves, 
        :math:`L` and :math:`R`, defined as follows:

        .. math::
            L = {(x, y) | x < 0 \lor x = 0 \land y < 0}

            R = {(x, y) | x > 0 \lor x = 0 \land y > 0}

        That is, :math:`R` contains all points with argument in the half-closed
        interval :math:`\\left[-\\frac{\pi}{2},\\frac{\pi}{2}\\right)`, and :math:`L` contains all 
        others.  Note that with these definitions, :math:`L` and :math:`R` are both 
        convex: a line segment between two points in :math:`R` lies entirely 
        in :math:`R`, and similarly for :math:`L`.  In particular, a line segment 
        between two points can only pass through the origin if one of those points is 
        in :math:`L` and the other in :math:`R`. Now, we follow the edges of the 
        polygon, keeping track of how many times we move between :math:`L` 
        and :math:`R`.  For each move from :math:`L` to :math:`R` (or vice versa),
        we also need to compute whether the edge passes *above* or *below* the origin,
        to compute its contribution to the total winding number.  From the comment
        above, we can safely ignore all edges that lie entirely within either :math:`L` 
        or :math:`R`.

```

coxeter/shapes/polygon.py
```diff
@@ -667,12 +666,50 @@ def compute_form_factor_amplitude(self, q, density=1.0):  # noqa: D102
...
+        only pass through the origin if one of those points is in L and the other in R.
+
+        Now the idea is that we follow the edges of the polygon, keeping track of how
+        many times we move between L and R.  For each move from L to R (or vice versa),
+        we also need to compute whether the edge passes *above* or *below* the origin,
+        to compute its contribution to the total winding number.  From the comment
+        above, we can safely ignore all edges that lie entirely within either L or R.
+
+        This implementation is a numpy vectorized version of the algorithm described.
+
```
janbridley, line 706:
```suggestion
        """Simple point-in-polygon algorithm based on winding number.

        The code in this function is based on implementation in :cite:`Dickinson2019` 
        which is licensed under the BSD-3 license. 
        
        Given a closed, possibly non-simple polygon described as a list of vertices in 
        :math:`\mathbb{R}^2` and a point that doesn't lie directly on the path of the 
        polygon, we'd like to compute the winding number of the polygon around the 
        point. To achieve this, we place the point at the origin. Divide the remainder 
        of the plane (i.e., :math:`\mathbb{R}^2` minus the origin) into two halves, 
        :math:`L` and :math:`R`, defined as follows:

        .. math::
            L = {(x, y) | x < 0 \lor x = 0 \land y < 0}

            R = {(x, y) | x > 0 \lor x = 0 \land y > 0}

        That is, :math:`R` contains all points with argument in the half-closed
        interval :math:`\\left[-\\frac{\pi}{2},\\frac{\pi}{2}\\right)`, and :math:`L` 
        contains all others.  Note that with these definitions, :math:`L` and :math:`R` 
        are both convex: a line segment between two points in :math:`R` lies entirely 
        in :math:`R`, and similarly for :math:`L`.  In particular, a line segment 
        between two points can only pass through the origin if one of those points is 
        in :math:`L` and the other in :math:`R`. Now, we follow the edges of the 
        polygon, keeping track of how many times we move between :math:`L` 
        and :math:`R`.  For each move from :math:`L` to :math:`R` (or vice versa),
        we also need to compute whether the edge passes *above* or *below* the origin,
        to compute its contribution to the total winding number.  From the comment
        above, we can safely ignore all edges that lie entirely within either :math:`L` 
        or :math:`R`.

janbridley: I have made some reviews to the text of the docstring, and have reformatted the math to properly use sphinx latex math. Feel free to make any other changes to this text if wanted - this could certainly be a starting point.

janbridley (changes requested):
I added some comments regarding the doctoring above, but the code looks great.

janbridley (approved):

Issue #44, opened 2020-01-19
Improved caching


vyasr:
### Feature description

Right now all properties are computed every time they are called. We should implement a caching mechanism by which the computed values are stored the first time they are computed.

### Proposed solution

We should implement a decorator, e.g. `cached_property`, that stores computed properties into a global dict cache if they are not found in the cache. We should also implement a second decorator `invalidates_cached_properties(property_names)` that removes data from the cache whenever a function is run that invalidates some property (e.g. rescaling the area of a polygon invalidates the moment of inertia calculation).

PR #188, opened 2023-08-08
Performance-focused rewrite of the ConvexPolyhedron class

janbridley:
Prior to this PR, the ConvexPolyhedron class used Scipy's (Qhull) convex hull to compute faces of the solid. The shape was then initialized in the same manner as a standard Polyhedron. This rewrite stores simplices and other data from the convex hull, which can be used to accelerate many of the classes methods. Slow and commonly run functions have been rewritten for performance, and new and/or improved pytests have been written.

## Description
The new class builds a convex hull, then stores most of the data from that object. This data, especially the list of simplices, allow for numpy vectorization of many functions that required a for loop in the previous code. These simplices can be sorted to ensure math involving the curl/divergence theorems works properly, and can be merged based on common normals to build faces similar to the current merge_faces function. 

**The following methods have been reimplemented, and a brief description is given:**

* `find_equations`: vectorized using numpy
* `_surface_triangulation`: updated to use precalculated simplices, so polytri is not required
* `volume`, `surface_area`: reimplemented to use stored volume data for the calculations. 
* `centroid`: rewritten to calculate from a list of simplices. Vectorized using numpy, and computes solutions using (`_centroid_from_triangulated_surface`)
* `_compute_inertia_tensor`: fully rewritten for performance, using an [algorithm](https://dl.acm.org/doi/10.1145/355873.355885) that works with tetrahedral decompositions that include negative-volume tetrahedra. It is also significantly faster.
* `_rescale`: updated to properly scale all stored properties. Also recalculates the centroid to ensure proper behavior.
* `get_face_area`: vectorized with numpy

**The following new methods have been added:**

* `_combine simplices`: faster alternative to merge faces, but only works on simplices
* `_find_simplex_equations`: similar to _find_equations, but works on every simplex rather than every face
* `_find_coplanar_simplices`: helper function that finds simplices with the same plane equation (within a tolerance). 
* `_calculate_signed_volume`: faster method of computing volume via a tetrahedral decomposition. Used as an internal check that volume is stable when modifying the shape
* `_find_face_centroids`: used with the new face_centroids property. Quickly finds the center of mass for each of the polygonal faces

**The following properties have been added:**
* `simplices`
* `equations`
* `face_centroids`

## Motivation and Context
Creation of a ConvexPolyhedron object takes on the order of 1e-2s - this is fine when generating small numbers of shapes, but can become a significant hindrance when generating tens or hundreds of thousands of shapes. In addition, many commonly used methods (especially volume, centroid, and inertia tensor) took on the order of 1e-3s and did not cache their data. These changes address these performance issues, and add a wide range of private variables which can be used to more quickly calculate geometric properties that have not yet been implemented. Benchmarks are included below, but in general this rewrite provides a 10x-100x performance improvement on improved methods. For properties with caching, values are only recalculated when the shape is modified and access the data takes on the order of 1e-8s

## Benchmarks
_Note that tests were performed in much more depth than shown below - these are simply a subset of the total_

#### Cube (V=8, F=6, E=12)
```python
cube = PlatonicFamily.get_shape("Cube")

%timeit -r 10 -n 10_000 ConvexPolyhedron(cube.vertices)
> 301 us ± 34.2 us per shape

%timeit -r 10 -n 100_000 cube._calculate_signed_volume()
> 7.69 us ± 239 ns per shape

%timeit -r 10 -n 100_000 cube._centroid_from_triangulated_surface()
> 25.7 us ± 676 ns per shape

%timeit -r 10 -n 100_000 cube.inertia_tensor
> 111 us ± 2.74 us per shape
```

#### Metabidiminished rhombicosidodecahedron (V=50, F=42, E=90)
```python
mbri = JohnsonFamily.get_shape("Metabidiminished Rhombicosidodecahedron")

%timeit -r 10 -n 10_000 ConvexPolyhedron(mbri.vertices)
> 1.81 ms ± 315 us per shape

%timeit -r 10 -n 100_000 mbri._calculate_signed_volume()
> 20.6 us ± 2.38 us per shape

%timeit -r 10 -n 100_000 mbri._centroid_from_triangulated_surface()
> 32.2 us ± 1.96 us per shape

%timeit -r 10 -n 100_000 mbri.inertia_tensor
> 194 us ± 9.65 us per shape
```

#### Cube (V=8, F=6, E=12)
```python
cube = PlatonicFamily.get_shape("Cube")

%timeit -r 10 -n 10_000 ConvexPolyhedron(cube.vertices)
> 5.22 ms ± 169 us per shape

%timeit -r 10 -n 100_000 cube.volume
> 1.38 ms ± 22.8 us per shape

%timeit -r 10 -n 100_000 cube.centroid
> 946 us ± 37.4 us per shape

%timeit -r 10 -n 100_000 cube.inertia_tensor
> 5.32 ms ± 478 us per shape
```

#### Metabidiminished rhombicosidodecahedron (V=50, F=42, E=90)
```python
mbri = JohnsonFamily.get_shape("Metabidiminished Rhombicosidodecahedron")

%timeit -r 10 -n 1_0000 ConvexPolyhedron(mbri.vertices)
> 38.1 ms ± 1.05 ms per shape

%timeit -r 10 -n 10_000 mbri.volume
> 10.2 ms ± 328 us per shape

%timeit -r 10 -n 10_000 mbri.centroid
> 8.34 ms ± 829 us per shape

%timeit -r 10 -n 10_000 mbri.inertia_tensor
> 34.5 ms ± 1 ms per shape
```

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


janbridley:
[0cf79cb](https://github.com/glotzerlab/coxeter/pull/188/commits/0cf79cb9b78ad40b6eaa4fe2cb02183fff490c40) is failing because it is dependent on #171 . Once that PR is merged, tests will pass

bdice (commented):
This is a pretty significant change for the `ConvexPolyhedron` class. I'm happy to see you were able to make it faster, but there are a lot of design questions here (such as whether it is a good idea to add a flag like `fast=True` to the constructor, which makes this class behave quite differently than the rest of `coxeter`). Generally, `coxeter` was designed to provide a common framework for representing geometry classes (which we previously lacked in the Glotzer group), and speed was not the central goal. This design choice is reflected in its implementation (aiming for readability and simplicity of implementation over speed in most cases).

Unfortunately I don't have bandwidth to review this properly in the next 2 weeks. I'd encourage seeking reviews from other maintainers/users of the coxeter package.

janbridley:
> there are a lot of design questions here (such as whether it is a good idea to add a flag like fast=True to the constructor, which makes this class behave quite differently than  the rest of coxeter). Generally, coxeter was designed to provide a common framework for  representing geometry classes (which we previously lacked in the Glotzer group), and speed was not the central goal. This design choice is reflected in its implementation (aiming for readability and simplicity of implementation over speed in most cases).


Totally valid to not have the bandwidth this week - I looped in @DomFijan for this. 

In terms of the scope of coxeter and the differences in mindset between this code (and a few other recent PRs) as compared to the state of the package a year ago: coxeter was created as a common geometry framework, but it was not initially intended to be used beyond creating a few dozens of shapes at a time. However, the tools here are widely applicable and many glotzergroup members (and presumably a few other researchers) are using the package as a comprehensive computational geometry package. This includes use creating hundreds or thousands of shapes at a time, as well as using coxeter inside molecular dynamics scripts to easily compute shape properties. There is definitely a use case for faster algorithms, and I would argue that although simple iteration based code may be easier to read, the barrier to understanding reasonably optimized numpy code is not significantly higher than the alternative. Speed is not the primary purpose of coxeter, and I do not think it should be - but orders-of-magnitude performance improvements with no loss in accuracy are well worth the slight increase in complexity.


In terms of the `fast` flag in particular: although its effectiveness is somewhat limited in the current implementation, there are plans to support convex decompositions in the future which would benefit from being able to use some or all of the ConvexPolyhedron code without a large number of extra calculations being performed on each segment. That being said, I am fine removing the feature in this PR, as it is not particularly difficult to implement.

janbridley:
I have marked this PR as a draft temporarily, as there is a significant bug when creating a ConvexPolyhedron with a non-convex input set. A failure in this case is fairly reasonable, but the error should be (a) more predictable and (b) well-documented.

### Current behavior:
```python
>>> verts
array([[-0.5,  0.5, -0.5],
       [-0.5,  0.5,  0.5],
       [ 0.5, -0.5, -0.5],
       [-0.5, -0.5, -0.5],
       [ 0.5,  0.5,  0.5],
       [-0.5, -0.5,  0.5],
       [ 0.5,  0.5, -0.5],
       [ 0. ,  0. ,  0. ],
       [ 0.5, -0.5,  0.5]])
>>> coxeter.shapes.ConvexPolyhedron(verts)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/$USER/coxeter/coxeter/shapes/convex_polyhedron.py", line 114, in __init__
    self._sort_simplices()
  File "/Users/$USER/coxeter/coxeter/shapes/convex_polyhedron.py", line 205, in _sort_simplices
    vertices = self._vertices[self._simplices]
               ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
IndexError: index 8 is out of bounds for axis 0 with size 8
```

When a non convex set is used as an input, vertices that don't contribute to the hull are removed from self._vertices. However - the _indices_ to those vertices stored in self._simplices are not updated, raising this error.

janbridley:
The class has been updated to raise a ValueError when a nonconvex set is provided. This error will also return a list of internal (or coplanar, degenerate, etc) points for easier debugging.

tests/conftest.py
```diff
@@ -37,6 +39,17 @@ def combine_marks(*marks):
...
+def is_not_ci():
+    if (
+        os.getenv("CI", "false") == "true"
+        or os.getenv("GITHUB_ACTIONS", "false") == "true"
+    ):
+        return False
+    else:
+        return True
+
+
```
DomFijan, line 52: You should remove the circle ci folder and file. You shouldnt need this code after that I think

.circleci/config.yml
```diff
@@ -0,0 +1,93 @@
...
+      - test-39
+      - test-38
+  deploy:
+    jobs:
+      - pypi_wheels:
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
```
DomFijan, line 93: You should remove this whole file

Credits.rst
```diff
@@ -113,6 +113,15 @@ Jen Bradley
...
 * Added edges, edge_vectors, edge_lengths, and num_edges methods.
+* Reimplemented ``find_equations``, ``_volume``, ``surface_area``, ``centroid``,
+  ``_compute_inertia_tensor``, ``rescale``, and ``get_face_area`` methods for convex
+  polyhedra using NumPy vectorized operations and polyhedron simplices.
+* Added the ``combine_simplices``, ``find_simplex_equations``, ``_find_face_centroids``,
+  ``find_coplanar_simplices``, ``_find_face_centroids``, and ``calculate_signed_volume``
+  methods to the ConvexPolyhedron class.
+* Added ``simplices``, ``equations``, and ``face_centroids`` properties to the
+  ConvexPolyhedron class.
+* Optimized pytest configurations for more efficient use of local and remote resources.
```
DomFijan, line 125: Need changelog entries as well?

DomFijan: Make sure you have a breaking changes category with API breaking changes.

tests/conftest.py
janbridley: This code is used to check the CI environment (GitHub actions) for pytest, which I think is useful - it allows us to define separate test behavior depending on the environment. There is probably a way to do this through GitHub actions with pytest marks, which I can look into if we want.

janbridley: This code is used to check the CI environment (GitHub actions) for pytest, which I think is useful - it allows us to define separate test behavior depending on the environment. There is probably a way to do this through GitHub actions with pytest marks, which I can look into if we want.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -4,47 +4,52 @@
...
     A convex polyhedron is defined as the convex hull of its vertices. The
-    class is a simple extension of :class:`~.Polyhedron` that builds the
-    faces from the simplices of the convex hull. This class also includes
-    various additional properties that can be used to characterize the
-    geometric features of the polyhedron.
+    class is an extension of :class:`~.Polyhedron` that builds the
+    faces from the simplices of the convex hull. Simplices are stored and class methods
+    are optimized to make use of the triangulation, as well as  special properties of
+    convex solids in three dimensions. This class also includes various additional
+    properties that can be used to characterize geometric features of the polyhedron.
```
DomFijan, line 27: These lines are slightly too long. They should be shorter than 80 characters I
think. At least most of the old docstrings are shorter than 80 chars.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,622 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+        self._simplex_equations = self._convex_hull.equations[:]
+        self._simplex_neighbors = self._convex_hull.neighbors[:]
+        self._volume = self._convex_hull.volume
+        self._area = self._convex_hull.area
+        self._maximal_extents = np.array(
+            [self._convex_hull.min_bound, self._convex_hull.max_bound]
+        )
+
+        # Clean up the result.
+        del self._convex_hull
```
DomFijan, line 137: Not sure if this logic makes sense. Wouldn't it be better to just pass
the convex hull to this function in the first place instead of first
assigninig it to `self._convex_hull` and deleting it after you are done
using it?

coxeter/shapes/convex_polyhedron.py
janbridley: I believe our black config is set to a line length of 88, but I can adjust the docstrings to meet the tighter spec.

coxeter/shapes/convex_polyhedron.py
janbridley: That is a much cleaner idea, thank you!

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,622 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+
+    def _combine_simplices(self, rounding: int = 15):
```
DomFijan, line 139: I'm not a huge fan of giving rounding as an int. I'd strongly advocate
this be passed as the actual precision we are looking to use ie. 1e-15
in this case.

coxeter/shapes/convex_polyhedron.py
janbridley: See [571d5b](https://github.com/glotzerlab/coxeter/pull/188/commits/571d5b51f37844eb5def545aa71285f91b24bf88)

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+
+        # Recompute simplex equations and centroid from the new triangulation.
+        self._find_simplex_equations()
+        self._centroid_from_triangulated_surface()
+
+    @property
+    def volume(self):
+        """float: Get or set the polyhedron's volume."""
+        return self._volume
+
```
DomFijan, line 264: These 3 functions have decently complicated logic and seem to be
important for a lot of the functionality of the class down the line.
Could you add a couple of tests for these? In case things break down it
would be very helpful to know where things fail. Especially if this is
not to time consuming to do. If coming up with tests is super hard maybe
its ok to leave without testing.

janbridley: I will do some thinking about it - the sorting isn't simple, but it should be testable at least for known cases.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+            equation_groups[equation_key].extend(self._simplices[i])
+
+        # Combine elements with the same plan equation and remove duplicate indices
+        ragged_faces = [
+            np.fromiter(set(group), np.int32) for group in equation_groups.values()
+        ]
+        self._faces = ragged_faces
+        self._equations = np.array(list(equation_groups.keys()))
+
+    def _find_coplanar_simplices(self, rounding: int = 15):
```
DomFijan, line 162: According to SciPy docs the qhull object already contains coplanar
indices in .coplanar? Is there a reason why you don't use that? See:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.ConvexHull.html

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+            equation_key = tuple(equation.round(rounding))
+            equation_groups[equation_key].append(i)
+        ragged_coplanar_indices = [
+            np.fromiter(set(group), np.int32) for group in equation_groups.values()
+        ]
+
+        self._coplanar_simplices = ragged_coplanar_indices
+
+    def _sort_simplices(self):
+        """Reorder simplices counterclockwise relatative to the plane they lie on.
```
DomFijan, line 188:
```suggestion
        """Reorder simplices counterclockwise relative to the plane they lie on.
```

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+        """Reorder simplices counterclockwise relative to the plane they lie on.
+
+        This does NOT change the *order* of simplices in the list.
+        """
+        # Get correct-quadrant angles about the simplex normal
+        vertices = self._vertices[self._simplices]
+
+        # Get the absolute angles of each vertex and fit to unit circle
+        angles = np.arctan2(vertices[:, 1], vertices[:, 0])
+        angles = np.mod(angles - angles[0], 2 * np.pi)
```
DomFijan, line 197: Is this necessary? This just changes which vertex you start at basically
right? I might be missing something. arctan2 codomain is -pi to pi.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -200,6 +799,36 @@ def maximal_centered_bounded_sphere(self):
...
+            >>> cube = coxeter.shapes.Polyhedron(
+            ...   vertices=cube.vertices, faces=cube.faces)
+            >>> import numpy as np
+            >>> assert np.isclose(cube.get_dihedral(1, 2), np.pi / 2.)
+
+        """
+        if b not in self.neighbors[a]:
+            raise ValueError("The two faces are not neighbors.")
+        n1, n2 = self._equations[[a, b], :3]
+        return np.arccos(np.dot(-n1, n2))
```
DomFijan, line 830: Dihedral should be on the whole circle (0 to 2pi). arccos codomain is
[0,pi]. I think you want arctan2 here.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+                   [ 1., -1.,  1.],
+                   [ 1.,  1., -1.],
+                   [ 1., -1., -1.],
+                   [-1.,  1.,  1.],
+                   [-1., -1.,  1.],
+                   [-1.,  1., -1.],
+                   [-1., -1., -1.]])
+
+        """
+        principal_moments, principal_axes = np.linalg.eigh(self.inertia_tensor)
```
DomFijan, line 686: If we dont use principal moments you can always do _ instead.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+
+        Returns
+        -------
+            float: Signed volume of the polyhedron.
+        """
+        signed_volume = np.sum(np.linalg.det(self._vertices[self._simplices]) / 6)
+        self._volume = abs(signed_volume)
+        return signed_volume
+
+    def get_face_area(self, face=None):
```
DomFijan, line 317: can we have some tests for this fn as well?

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+        self._vertices *= scale_factor
+        self._equations[:, 3] *= scale_factor
+        self._simplex_equations[:, 3] *= scale_factor
+        self._volume = self._volume * scale_factor**3
+        self._area = self._area * scale_factor**2
+
+        # Recalculate centroid of shape
+        self._centroid_from_triangulated_surface()
+
+    def _calculate_signed_volume(self):
```
DomFijan, line 302: Is it useful to have signed volume available to the user stored in a
variable? Or expose this function?

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+            # Calculate distances
+            distances = np.linalg.norm(vertices, axis=1)
+
+            # Create a tuple of distances and angles to use for lexicographical sorting
+            vert_order = np.lexsort((distances, angles))
+
+            # Apply reordering to every simplex
+            sorted_faces.append(face[vert_order])
+
+        self._faces = sorted_faces
```
DomFijan, line 522: Would it maybe make sense for faces to be convexPolygon-s instead of
just array of vertices? There might be some comp overhead to creating
them though.

DomFijan: Also simplices.

DomFijan (changes requested):
I've left some comments that we should discuss. Some methods are lacking tests that should be added. And some minor tweaks are suggested.

coxeter/shapes/convex_polyhedron.py
janbridley: There is a pretty significant performance overhead to doing this - furthermore, much of the array math breaks down. This would be another good use case for a "collection of shapes" object, but I don't think it makes sense here

coxeter/shapes/convex_polyhedron.py
janbridley: We could certainly expose the quantity if you think it is valuable! We already compute the quantity on initialization so that should be simple

coxeter/shapes/convex_polyhedron.py
janbridley: `shapes/Polyhedron` currently uses arccos, which I copied here as the initial code had changes to the method. I can remove this function and fall back to the (identical) code from  the nonconvex class, but we should take a look at the implementation offline

coxeter/shapes/convex_polyhedron.py
janbridley: Coplanar points in hull are not the contributors to coplanar simplices (as needed here). Rather, they are points that lie "on" a facet but cannot be triangulated due to numerical imprecision - for example, a point at [0,0,1-ε] on the face of a cube.

coxeter/shapes/convex_polyhedron.py
```diff
@@ -71,22 +76,616 @@ class is a simple extension of :class:`~.Polyhedron` that builds the
...
+    @property
+    def centroid(self):
+        """:math:`(3, )` :class:`numpy.ndarray` of float: Get or set the center of mass.
+
+        The centroid is calculated using the curl theorem over the surface simplices.
+        """
+        return self._centroid
+
+    @centroid.setter
+    def centroid(self, value):
```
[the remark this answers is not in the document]
janbridley: Setting the centroid takes ~130μs for a cube, and about 200 μs for a larger shape (metabidiminished rhombicosidodecahedron, as above). If we can find a clean pythonic way to achieve this I think that's a great idea - I will do some research

coxeter/shapes/convex_polyhedron.py
janbridley: Pytests for ConvexPolyhedron pass with or without the shift, but the current code mirrors the behavior of `shapes/_reorder_verts`. Removing the rescaling from ConvexPolygon causes the test `test_polygon.p::test_nonregular_convex_polygon_distance_to_surface_unit_area_ngon` to fail. It seems that reordering the vertices is somehow required to ensure the resulting polygon is simple. On a first glance I can't figure out why this would be the case, but I think we should explore further because the behavior is quite odd. This also highlights that we should add some 3D test cases with non regular faces - the current `named_solids_mark` does not include anything more irregular than Johnson solids

coxeter/shapes/convex_polyhedron.py
janbridley: See [9aa4a59](https://github.com/glotzerlab/coxeter/pull/188/commits/9aa4a59d2422f11edd8392d3ca036be33820a1bb)

coxeter/shapes/convex_polyhedron.py
janbridley: Do you have any thoughts on how this could be implemented? My current approach hashes the plane equation which makes comparison-based approaches trickier. I could try and reimplement this in another way, but I have not found a non-hashing based approach that is comparable in performance.

coxeter/shapes/convex_polyhedron.py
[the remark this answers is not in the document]
janbridley: I think that this is somewhat similar to the "fast" flag - logging which values need to be updated when other properties are changed is a pretty major change, and should perhaps be left for its own PR (see #44)

coxeter/shapes/convex_polyhedron.py
janbridley: Working on it - I caught a bug so it will be a few

janbridley: [4fe9d88](https://github.com/glotzerlab/coxeter/pull/188/commits/4fe9d8846b26b755bc8adbdbb66828c49513404e) addresses issues with the `_find_coplanar_simplices` method that were causing errors with get_face_area. The new method takes a floating point tolerance (rather than an integer number of decimal places), and combines all of the logic into the `_combine_simplices` method, as there was significant overlap in functionality. This addresses several of the comments given above.

coxeter/shapes/convex_polyhedron.py
janbridley: I solved this in [4fe9d88](https://github.com/glotzerlab/coxeter/pull/188/commits/4fe9d8846b26b755bc8adbdbb66828c49513404e) - tolerances are now done to 2e-15, and floating point values can be provided

coxeter/shapes/convex_polyhedron.py
janbridley: I would love to test this method directly, but worry that doing so would be complicated enough to warrant a separate PR. There are many unique solutions to sorting the simplices, and the results will depend on both the ConvexHull's initial simplex ordering and the ordering of the points of the input set.

However, the correct ordering is tested indirectly by computations of the centroid and intertia tensor. If the sorting is incorrect, those values will evaluate to nonsensical quantities and will be picked up by several tests: the monte carlo inertia tensor calculation and centroid setter tests should prove particularly good for this, although we have many tests for those properties.

The best solution would be to enumerate all possible orderings for a tetrahedron and/or octahedron and test against that, but for now I think we are safe to leave the method as-is.

DomFijan: @janbridley  is #211  related to this?

janbridley: They are unrelated - they are both artifacts of SciPy's `ConvexHull`, but result from different behaviors

janbridley:
@tcmoore3 Any other feedback? If not I will merge in the morning and start working on a release

tcmoore3 (approved):

PR #189, opened 2023-08-10
Optimize point in polyhedron

DomFijan:
Speed up point in polygon using numpy operations. 

## Description
I have moved most of explicit for loops into numpy optimized operation. This is a breaking change because points on boundary are no longer considered inside.
There is potentially more speedup to be had, but this involves optimizing polytri implementation. Probably for a different PR.

## Motivation and Context
Speed for Truncated Dodecahedron test for 1000 points:
master implementation `234 ms ± 49.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`
this branch `23.5 ms ± 2.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:


janbridley (approved):
Looks great! Thank you for adding this

coxeter/shapes/polyhedron.py
```diff
@@ -820,38 +823,75 @@ def is_inside(self, points):
...
+        # triangle vertices
+        v1 = self.vertices[triangles[:, 0]]
+        v2 = self.vertices[triangles[:, 1]]
+        v3 = self.vertices[triangles[:, 2]]
+        v1_expanded = v1[:, None, :]
+        v2_expanded = v2[:, None, :]
+        v3_expanded = v3[:, None, :]
+        points_expanded = np.tile(points, (v1.shape[0], 1, 1))
+
+        diff_x_v1 = v1_expanded[..., 0] - points_expanded[..., 0]
```
bdice, line 850: If possible, I would avoid breaking this into components. I understand it might have been done for readability (giving the names x, y, z) but accessing array indices 0, 1, 2 will probably make it easier to vectorize the computations here?

coxeter/shapes/polyhedron.py
```diff
@@ -820,38 +823,75 @@ def is_inside(self, points):
...
+
+        v1sign = sign_or(np.sign(diff_x_v1), np.sign(diff_y_v1), np.sign(diff_z_v1))
+        v2sign = sign_or(np.sign(diff_x_v2), np.sign(diff_y_v2), np.sign(diff_z_v2))
+        v3sign = sign_or(np.sign(diff_x_v3), np.sign(diff_y_v3), np.sign(diff_z_v3))
+
+        mask12 = v1sign != v2sign
+        mask23 = v2sign != v3sign
+        mask31 = v3sign != v1sign
+
+        term_1_1 = diff_y_v1 * diff_x_v2 - diff_x_v1 * diff_y_v2
```
bdice, line 871: If we rewrote this with vectors and re-indexed, this could be written as something like:
```python
def cross(vecs1, vecs2, i, j):
    return ...

term_0_0 = cross(diff_v1, diff_v2, 0, 0)
...
```
or even combine the cross terms inside that function definition and just write `term_0 = cross(diff_v1, diff_v2)`.

coxeter/shapes/polyhedron.py
```diff
@@ -820,38 +823,75 @@ def is_inside(self, points):
...
+            [
+                [vertex_to_index[tuple(v)] for v in triangle]
+                for triangle in self._surface_triangulation()
+            ]
+        )
+        points = np.atleast_2d(points)
 
-        winding_number_calculator = PolyInside(triangles, self.vertices)
+        # triangle vertices
+        v1 = self.vertices[triangles[:, 0]]
```
bdice, line 842: I would suggest re-indexing `v1, v2, v3` to `v0, v1, v2` and then using indices to indicate array components.

coxeter/shapes/polyhedron.py
```diff
@@ -820,38 +823,75 @@ def is_inside(self, points):
...
+        v1sign = sign_or(np.sign(diff_x_v1), np.sign(diff_y_v1), np.sign(diff_z_v1))
+        v2sign = sign_or(np.sign(diff_x_v2), np.sign(diff_y_v2), np.sign(diff_z_v2))
+        v3sign = sign_or(np.sign(diff_x_v3), np.sign(diff_y_v3), np.sign(diff_z_v3))
+
+        mask12 = v1sign != v2sign
+        mask23 = v2sign != v3sign
+        mask31 = v3sign != v1sign
+
+        term_1_1 = diff_y_v1 * diff_x_v2 - diff_x_v1 * diff_y_v2
+        term_2_1 = diff_z_v1 * diff_x_v2 - diff_x_v1 * diff_z_v2
```
bdice, line 872: This cycle seems a little funny to me. It seems like normally we should see a cycle like `xy, yz, zx` or `yx, zy, xz` but this has `yx, zx, zy`. The ordering of components (and thus sign conventions) doesn't seem quite right. Maybe there's an implied negative sign someplace that I'm not accounting for.

tests/test_polyhedron.py
```diff
@@ -739,12 +739,19 @@ def test_get_set_minimal_centered_bounding_sphere_radius(poly):
 )
 def test_is_inside(cube):
     assert cube.is_inside(cube.center)
+    import sys
 
-    @given(floats(0, 1), floats(0, 1), floats(0, 1))
-    def testfun(x, y, z):
+    limit = sys.float_info.min
```
bdice, line 744: Let's use numpy instead and remove the `sys` import. This is also clearer (smallest normal is a more precise term than "min"). On my system these appear to give the same value.
```suggestion
    limit = np.finfo(np.float64).smallest_normal
```

bdice:
I left some comments, suggestions, and questions. Please feel free to merge if you need to move forward, I may not be able to review again for a bit.

tests/test_polyhedron.py
DomFijan: I will also update point in polygon test to use the same logic.

coxeter/shapes/polyhedron.py
DomFijan: all 1,2,3 indecies have been changed to 0,1,2 now

ChangeLog.rst
```diff
@@ -15,6 +15,7 @@ Changed
 - The minimum required Python version is now 3.8.
 - The minimum required NumPy version is now 1.19.
 - Speed up point in polygon using NumPy.
+- Speed up point in polyhedron using NumPy.
```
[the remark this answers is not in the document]
DomFijan: I'm working on finishing this off now. 
The breaking change also applies for polygon. And putting the two breaking changes in a new release would make a lot of sense.

coxeter/shapes/polyhedron.py
DomFijan: I think its just a cross product. Terms term_1_1 , term_2_1 and term_3_1 are vectors of a cross product of v1-point and v2-point. I think the confusing part is in how xyz vs v1,v2,v3 is defined and how the whole expression is layed out. Admittedly this is confusing. Refactoring using a cross function will help readability.

coxeter/shapes/polyhedron.py
DomFijan: I actually did it for speed. Had the exact same thought, but for some reason code slows down for about 50%. I'm not sure why. Using the same test as in the PR description I get following results
master implementation
`190 ms ± 21.7 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)`
Proposed solution
`19 ms ± 1.65 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)`
diffs vectorized
`28.7 ms ± 1.01 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)`

diffs vectorized code:
  ```
    triangles_expanded = np.expand_dims(vertices[triangles] , axis=1)  # Shape (M, 1, 3, 3)
    points_expanded = np.expand_dims(np.expand_dims(points, axis=0), axis=2)  # Shape (1, N, 1, 3)


    # Compute differences, shape (M, N, 3, 3)
    differences = triangles_expanded - points_expanded

    # Compute signs, shape (M, N, 3, 3)
    signs = np.sign(differences)

    # Compute vertex signs using OR-like operation, shape (M, N, 3)
    vertex_signs_x = sign_or(signs[..., 0, 0], signs[..., 0, 1], signs[..., 0, 2])
    vertex_signs_y = sign_or(signs[..., 1, 0], signs[..., 1, 1], signs[..., 1, 2])
    vertex_signs_z = sign_or(signs[..., 2, 0], signs[..., 2, 1], signs[..., 2, 2])
```

File: coxeter/shapes/polyhedron.py (hunk 4 above)
Comment (DomFijan) [reply coxeter/shapes/polyhedron.py]: Actually you are right about something being off. Something is different here, like everything is sort of "shifted" around and some terms have an implicit minus or something like that. Maybe this is not a "classic" cross product. The way I made this is by following the original implementation formulas.

Comment (DomFijan) [reply coxeter/shapes/polyhedron.py]: I think the edge sign and triangle sign use the terms in different ways - with and without a minus. And I adapted this from the edge sign function where it is originally computed with this "implicit minus".

File: coxeter/shapes/polyhedron.py (hunk 2 above)
Comment (DomFijan) [reply coxeter/shapes/polyhedron.py]: I have now tested a version with the suggested change and it seems to yield a slightly faster code, although its within error of the proposed implementation. I think it makes sense to change the calculation as proposed. The results are as follows:
```
master implementation
184 ms ± 12.6 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
Proposed solution
20 ms ± 2.35 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
cross term vectorized
19.1 ms ± 1.18 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

```
cross term vectorized:
    def compute_cross(diff_i, diff_j):
        """
        Helper function to compute the terms.
        """
        term_1 = diff_i[1] * diff_j[0] - diff_i[0] * diff_j[1]
        term_2 = diff_i[2] * diff_j[0] - diff_i[0] * diff_j[2]
        term_3 = diff_i[2] * diff_j[1] - diff_i[1] * diff_j[2]
        return term_1, term_2, term_3

    mask12 = v1sign != v2sign
    mask23 = v2sign != v3sign
    mask31 = v3sign != v1sign

    term1 = compute_cross((diff_x_v1, diff_y_v1, diff_z_v1), (diff_x_v2, diff_y_v2, diff_z_v2))
    term2 = compute_cross((diff_x_v2, diff_y_v2, diff_z_v2), (diff_x_v3, diff_y_v3, diff_z_v3))
    term3 = compute_cross((diff_x_v3, diff_y_v3, diff_z_v3), (diff_x_v1, diff_y_v1, diff_z_v1))

    edge1 = sign_or(*[np.sign(term) for term in term1])
    edge2 = sign_or(*[np.sign(term) for term in term2])
    edge3 = sign_or(*[np.sign(term) for term in term3])

    triangle_sign = np.sign(-term1[0] * diff_z_v3 - term2[0] * diff_z_v1 - term3[0] * diff_z_v2)
    
```

**Comment by DomFijan** [comment]
@bdice let me know if you want to/have time to take another final look at this.

**Review by janbridley — APPROVED** [review]
Looks very good - thank you for this!

File: ChangeLog.rst (hunk 6 above)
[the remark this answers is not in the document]
Comment (bdice) [reply ChangeLog.rst:18]: I'd recommend changing the header above this to `v0.7.0` instead of `v0.6.2` to align with this intent.

File: coxeter/shapes/polyhedron.py [hunk 7]
```diff
@@ -857,38 +860,81 @@ def is_inside(self, points):
...
+
+        v0sign = sign_or(np.sign(diff_x_v0), np.sign(diff_y_v0), np.sign(diff_z_v0))
+        v1sign = sign_or(np.sign(diff_x_v1), np.sign(diff_y_v1), np.sign(diff_z_v1))
+        v2sign = sign_or(np.sign(diff_x_v2), np.sign(diff_y_v2), np.sign(diff_z_v2))
+
+        mask01 = v0sign != v1sign
+        mask12 = v1sign != v2sign
+        mask20 = v2sign != v0sign
+
+        def compute_cross(diff_i, diff_j):
```
bdice, line 908: Can you write this with `np.cross` for clarity (and probably performance)? https://numpy.org/doc/stable/reference/generated/numpy.cross.html

It looks pretty close up to some relabeling of the values. Seems like that gives `[-term2, term1, -term0]` relative to the terms you want.

bdice (approved):
A few small comments -- this is dramatically improved, please merge when you're happy with it.

coxeter/shapes/polyhedron.py
DomFijan: The correct expression is:
` np.moveaxis(-np.cross(diffi, diffj, axis=2)[..., [2,1,0]], -1, 0)`
However this slows down by a factor of two:
```
master implementation
359 ms ± 16.6 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
First proposed solution
38.3 ms ± 1.8 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
current solution vectorized
37.4 ms ± 4.5 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
using numpy for cross term
62.7 ms ± 2.47 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```
It probably has to do with how slicing works. It might just be faster to have precomputed sliced arrays (just like in the case of the first comment that slowed down by a factor of 2).

DomFijan: I'll add some comments to the code re this.

PR #177, opened 2023-02-13
Added new shape families for Archimedean, Catalan, Johnson, and other solids.

janbridley:
## Description
Expanded on coxeter.families with methods and data for several families of common convex polyhedra:
* ``ArchimedeanFamily``: The 13 Archimedean solids.
* ``CatalanFamily``: The 13 Archimedean duals.
* ``JohnsonFamily``: The 92 Johnson solids.
* ``PrismAntiprismFamily``: The 16 n-gonal prisms and antiprisms ranging from n=3 to n=10.
* ``PyramidDipyramidFamily``: The 6 equilateral pyramids and dipyramids.

Each new family is queryable with ``get_shape("Shape Name")`` and returns a ``ConvexPolyhedron`` normalized to volume of 1, just as ``PlatonicFamily`` currently does. Vertex data for all new families is stored as a float64, and stored data for the current ``PlatonicFamily`` has been updated to match that precision. This change fixes intermittent bugs that arose from lack of numerical precision. New pytests have been added to verify the new families behave as expected. 

## Motivation and Context
Querying polyhedra other than Platonic solids could be relatively difficult. This release streamlines the process and allows for simple access to a greatly increased number of convex solids.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:
- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/coxeter/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/coxeter/blob/master/ContributorAgreement.md).
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.
- [x] I have updated the [changelog](https://github.com/glotzerlab/coxeter/blob/master/ChangeLog.txt) and the [credits](https://github.com/glotzerlab/coxeter/blob/master/doc/source/credits.rst).

PR #191, opened 2023-08-30
Fixed documentation for common families

janbridley:
## Description
A few of the families defined in #177 have messy documentation that result from missing line breaks. This has been addressed, and the short docstrings (which display in the Classes table on readthedocs) have been made more uniform.

## Types of Changes
- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:
<sup>These changes fix a PR that I authored, so I have not modified the changelog.</sup>

### DIFF
--- a/coxeter/families/common.py
+++ b/coxeter/families/common.py
@@ -138,7 +138,7 @@ PyramidDipyramidFamily = TabulatedGSDShapeFamily.from_json_file(
     docstring="""The family of regular equilateral pyramids and dipyramids (6 total).
 
     The following parameters are required by this class:
-    
+
     - name: The name of the pyramid or dipyramid. Options are "Triangular Pyramid", \
             "Square Pyramid", "Pentagonal Pyramid", "Triangular Dipyramid", \
             "Square Dipyramid", and "Pentagonal Dipyramid".
--- a/coxeter/families/common.py
+++ b/coxeter/families/common.py
@@ -75,7 +75,7 @@ class RegularNGonFamily(ShapeFamily):
 PlatonicFamily = TabulatedGSDShapeFamily.from_json_file(
     os.path.join(_DATA_FOLDER, "platonic.json"),
     classname="PlatonicFamily",
-    docstring="""The family of Platonic solids.
+    docstring="""The family of Platonic solids (5 total).
 
 The following parameters are required by this class:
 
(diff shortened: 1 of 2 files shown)


coxeter/families/common.py
```diff
@@ -117,23 +119,25 @@ def make_vertices(cls, n):
...
 )
 
-
 PyramidDipyramidFamily = TabulatedGSDShapeFamily.from_json_file(
     os.path.join(_DATA_FOLDER, "pyramid_dipyramid.json"),
     classname="PyramidDipyramidFamily",
     docstring="""The family of regular equilateral pyramids and dipyramids (6 total).
-The following parameters are required by this class:
+
+    The following parameters are required by this class:
```
DomFijan, line 140: This doesn't render properly. I think you want the following:
```suggestion
    The following parameters are required by this class:
    
```

coxeter/families/common.py
```diff
@@ -143,8 +147,10 @@ def make_vertices(cls, n):
 PrismAntiprismFamily = TabulatedGSDShapeFamily.from_json_file(
     os.path.join(_DATA_FOLDER, "prism_antiprism.json"),
     classname="PrismAntiprismFamily",
-    docstring="""The family of n-gonal prisms and antiprisms with n∈[3,10] (16 total).
-The following parameters are required by this class:
+    docstring="""The family of uniform n-prisms and n-antiprisms with n∈[3,10] \
+    (16 total).
+
+    The following parameters are required by this class:
```
DomFijan, line 153: Same as above:
```suggestion
    The following parameters are required by this class:
    
```

DomFijan (changes requested):
Looks good, Some of the name docs don't render properly (at least they are not in line with how others look). I think you are missing an empty line. Once this is fixed this is ready.

coxeter/families/common.py
janbridley: Looks like the body text is inconsistent for a few of the families. I will go through and fix that.

janbridley: See [cd06702](https://github.com/glotzerlab/coxeter/pull/191/commits/cd06702226511160d6de113bb30c5c2972adca9a)

DomFijan (approved):

PR #193, opened 2023-08-30
Migrated setup.py and setup.cfg to pyproject.toml

janbridley:
This change aims to bring the coxeter install files up to spec with PEP621. Both setup.py and setup.cfg have been replaced with minimal versions.

## Motivation and Context
setup.py install has been deprecated, meaning that a working (and complete) pyproject.toml is important to ensuring users can install from source. In addition, the installation documentation has been updated to ensure the proper install script (`pip install .[tests]`) is recommended to the user.

## Testing
- [x] pyproject.toml validity has been checked with pyproject.toml
- [x] `pip install .[tests]; pytest` works as expected
- [x] `pip install .[doc]; cd doc; make html` works as expected
- [x] Update docs
- [x] Test building wheels

#### Testing wheels
```bash
>python -m pip wheel .
Processing /Users/$USER/Documents/GitHub/coxeter
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
...
Building wheels for collected packages: coxeter
  Building wheel for coxeter (pyproject.toml) ... done
  Created wheel for coxeter: filename=coxeter-0.6.1-py3-none-any.whl size=158710 sha256=4726a0e8f33a894b0d7c3f09462c86f7a62b102a90aff49716886b0675bc9cfa
  Stored in directory: dir
Successfully built coxeter

>ls | grep whl
coxeter-0.6.1-py3-none-any.whl

>python -m pip install coxeter-0.6.1-py3-none-any.whl

Processing ./coxeter-0.6.1-py3-none-any.whl
Obtaining dependency information for...
Installing collected packages: numpy, scipy, rowan, coxeter
  ...
Successfully installed coxeter-0.6.1 numpy-1.25.2 rowan-1.3.0.post1 scipy-1.11.2
```

## Types of Changes
- [x] Documentation update
- [x] Bug fix
- [ ] New feature

_Optional requirements [test] has been renamed to [tests]._

## Checklist:


janbridley:
Wheel building is working, docs build, and tests pass. This is a big change and a second verification of those tests would be greatly appreciated - but it seems like just about everything is in order.

.DS_Store
bdice, line 1: Please remove `.DS_Store` and perhaps add it to the coxeter `.gitignore` (maybe align with what other projects do here). You can also prevent git from ever adding files like this if you create a file `~/.gitexcludes`. Mine looks like this:

```
.DS_Store
*.pyc
._*
**/.ipynb_checkpoints
*.swp
__pycache__/
*.egg-info*
.vscode
.hypothesis
```

README.rst
```diff
@@ -116,17 +124,21 @@ Building Documentation
...
 
 .. code:: bash
 
-   cd doc
-   pip install -r requirements.txt
+   pip install -r doc/requirements.txt
+   conda install -c conda-forge fresnel
+
+.. warning::
+   The `fresnel <https://fresnel.readthedocs.io/en/v0.13.5/>`_ package on conda forge must be used. The PyPI package *fresnel* is different and will not function properly.
```
bdice, line 135: Don't link to a specific version.
```suggestion
   The `fresnel <https://fresnel.readthedocs.io/>`_ package on conda-forge must be used. The PyPI package *fresnel* is different and will not function properly.
```

bdice (commented):
Very brief review comments attached. Generally looks fine.

doc/requirements.txt
```diff
@@ -2,8 +2,9 @@ numpy
 rowan>=1.2
 scipy
 jupyter
+sphinx==5.3.0
```
joaander, line 5: Why the hard pin?

joaander (commented):
Thanks! This is a good improvement. I have one question.

doc/requirements.txt
janbridley: I pinned at 5.3.0 for testing, but `make html` works for sphinx between 4.2.0 to 5.3.0. I will update requirements.txt to match!

.DS_Store
janbridley: Heard - not sure how this one slipped through, but I will remove it shortly.

doc/requirements.txt
joaander: There are many problems with upper limits, use them only when strictly necessary. See https://iscinumpy.dev/post/bound-version-constraints/ for details.

janbridley: Very interesting - I had not considered this. It seems this is a very relevant document, as the issue I was fixing here is more complex than I realized.

Long explanation of what seems to be happening:
_When attempting to install newer version of sphinx, there is a solver error `sphinx-rtd-theme NEWER_VERSION requires docutils<0.18` because sphinx 7.2.5 requires `docutils >0.18.1`. To get around this, pip must fall back to a (very) old version of sphinx-rtd-theme: 0.5.1, as shown above, which then requires an older version of sphinx (<=5.3.0) to function properly, despite the requirements claiming any version of sphinx will work.  This is further complicated because in `sphinx-rt-theme==1.2.0rc2`, the sphinx version was pinned between `sphinx>=6.0,<6.1`, and in `sphinx-rt-theme==1.3.0rc1` the required versions were changed to `sphinx>=7.1,<7.2`: so a newer version of sphinx-rtd-theme places a rather tight restriction on valid versions of sphinx to install. All this is to say, because of the conflicting requirements (especially the upper bounds for compatible sphinx version), a very old version of sphinx-rtd-theme was installed by default. _ 

Rather than optimizing sphinx versions around this, sphinx-rtd-theme versions should be optimized around sphinx. Adding the constraint `sphinx-rtd-theme>=1.3.0` seems to fix everything, as it is compatible with the current sphinx and seems to operate fine even with older versions (despite not officially having support for them).

joaander: That is more complicated than I was expecting. Adding requirements to *sphinx-rtd-theme* and not *sphinx* is the right solution here. When you switch from *sphinx-rtd-theme* to *furo*, then *furo* will be able to set the appropriate version bounds for *sphinx*`.

janbridley: Excellent, I will keep that in mind when I work on the _furo_ migration. Thank you for the resource above, it was extremely helpful in identifying the underlying issue here.

joaander (approved):
This looks good. Thanks!

PR #196, opened 2023-09-18
Added `edge_lengths` property

janbridley:
## Description
Added edge_lengths property to the Polyhedron class.

## Motivation and Context
Although this is a single line of code, I use it extremely frequently (as it has additional value in helping to determine the degree of uniformity of a solid). This should be a quick PR, and if it seems unnecessary I am happy to continue manually implementing this function.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


coxeter/shapes/polyhedron.py
```diff
@@ -390,6 +390,11 @@ def edge_vectors(self):
         """:class:`numpy.ndarray`: Get the polyhedron's edges as vectors."""
         return self.vertices[self.edges[:, 1]] - self.vertices[self.edges[:, 0]]
 
+    @property
+    def edge_lengths(self):
+        """:class:`numpy.ndarray`: Get the length of each edge of the polyhedron."""
```
DomFijan, line 395:
```suggestion
        """:class:`numpy.ndarray`: Get the length of each edge of the polyhedron in the same order as order of edges in :meth:`edge_vectors` ."""
```

DomFijan (changes requested):
This looks good. Lets make sure that order of edge lengths is well defined in the docs and this is good to go.

coxeter/shapes/polyhedron.py
janbridley: Should we add this ordering specification to edge_vectors as well? Both use the same ordering as :meth:`edges` which is the primary sorted array

DomFijan: Maybe. Is it super obvious what this is? If it is maybe this needs not be in the docstring, but if it is not then it should be for sure.

DomFijan (approved):

PR #200, opened 2023-09-19
Migrate CI to GitHub actions

DomFijan:
Migration from Circle CI will involve removing Circle CI files and replacing them with github actions files. Additionally, dependabot integration will be added.
 
## Description
Swap testing with github actions. 
Swap PyPI uploading upon tagging to github CI.
Re-do the testing approach to be compatible with dependabot usage.

## Motivation and Context
Github actions > Circle CI ?

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


tests/requirements.txt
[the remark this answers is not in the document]
joaander: Separate requirements files that are for *users* and for *CI*. They are used under different scenarios. Requirements that will be used as part of solving environments are best with lower bounds only (see https://iscinumpy.dev/post/bound-version-constraints/) and upper bounds only in very controlled circumstances.

For *CI*, however, we want exact pins. With no pin, rerunning CI on the *same commit* can pass one day and fail the next due to upstream releases. Using exact pins mitigates this risk.

Dependabot will periodically issue pull requests to update the exact pins. In this way, you can ensure, one pull request at a time, that each new upstream dependency release passes tests (and any needed changes are made) before updating the pin. See https://github.com/glotzerlab/gsd/pull/281 for an example of a dependabot pull request.

I often put CI specific requirements files in the `.github` directory.

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  run_tests:
+    name: Run tests on ubuntu-latest with Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8","3.9","3.10","3.11"]
+    steps:
+    - uses: actions/checkout@v3
```
joaander, line 27:
```suggestion
    - uses: actions/checkout@v4
```
Checkout is on v4 now. If you don't mind a few extra dependabot PRs, use `v4.0.0` to ensure repeatability.

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
```
joaander, line 29:
```suggestion
      uses: actions/setup-python@v4.7.0
```
[optional] (see discussion above)

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,46 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install -r tests/requirements.txt
+    - name: Install package
+      shell: bash -l {0}
```
joaander, line 36:
```suggestion
```
Bash login shells are only needed for conda environments.

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        python --version
+        python -c "import numpy; print('numpy', numpy.__version__)"
+        python -c "import scipy; print('scipy', scipy.__version__)"
+        python -m pip install .[tests] --progress-bar off
+    - name: Test with pytest
+      shell: bash -l {0}
```
joaander, line 43:
```suggestion
```
Same here.

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        python -m pip install pytest
```
joaander, line 45:
```suggestion
```
pytest is in the requirements file.

.github/workflows/PublishPyPI.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Build source distribution and wheels
+      run: |
+        python --version
+        python -m pip install --progress-bar off twine wheel setuptools
+        python setup.py sdist
```
joaander, line 22: I thought `python setup.py <anything>` was deprecated and removed. I use the [build](https://pypi.org/project/build/) in gsd to make the sdist. I think it can construct the wheel needed for coxeter as well (can test locally to be sure).

.github/workflows/PublishPyPI.yaml
```diff
@@ -0,0 +1,26 @@
...
+        python setup.py bdist_wheel
+    - name: Publish to PyPI
+      run: |
+        python -m twine upload --username __token__ --password ${{ secrets.PYPI_API_TOKEN }} dist/*
```
joaander, line 26: PyPI recommends trustless publishing going forward: See https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/. The advantage is that you do not need per-repository secrets. The *disadvantage* is that you accept the risk that all packages trustless publishing might be hacked simultaneously by one security breach.

.github/workflows/PublishPyPI.yaml
```diff
@@ -0,0 +1,26 @@
+name: Build and Publish to PyPI
+
+on:
+  push:
```
joaander, line 4: Test the wheel build step on pull requests as well. Skip the upload step when it is not a tag build (or upload to testpypi instead if desired). See gsd for an example: https://github.com/glotzerlab/gsd/blob/70f7c04349646dd997ae23a3f90aecfbb6590f65/.github/workflows/build_wheels.yml#L91-L97

.github/workflows/PublishPyPI.yaml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        python-version: '3.x'
+    - name: Build source distribution and wheels
+      run: |
+        python --version
+        python -m pip install --progress-bar off twine wheel setuptools
+        python setup.py sdist
+        python setup.py bdist_wheel
+    - name: Publish to PyPI
+      run: |
```
joaander, line 25: Use [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) instead of manual twine commands.

joaander (changes requested):
Thanks! I have suggestions for improvement.

.github/workflows/PublishPyPI.yaml
DomFijan: We discussed offline and decided to not use trustless publishing yet.

tests/requirements.txt
janbridley, line 1: These changes should probably go in a .github/test-requirements.txt - otherwise, normal users who `pip install .[tests]` will get the pinned packages

joaander (approved):
Thanks!

janbridley (approved):

PR #203, opened 2023-09-28
Switch linting to ruff

DomFijan:
Flake8 is slow and sometimes doesn't play well with black.

## Description
Ruff is a replacement for flake8 and many other linters. It is fast. Also works on jupyter notebooks. It's supposed to play well with black?

## Motivation and Context
Modernize our linting setup.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


pyproject.toml
```diff
@@ -64,11 +54,46 @@ exclude = '''
...
   "D107", # __init__ should be documented in class constructors.
+  "D203", # incompatible with D211
+  "D213", # incompatible with D212
   "D401", # This doesn't play nice with property docstrings of the form `type: ...`
   "D402", # This doesn't play nice with property docstrings of the form `type: ...`
-  "D403" # This doesn't play nice with property docstrings of the form `type: ...`
+  "D403", # This doesn't play nice with property docstrings of the form `type: ...`
+  "D404", # enables this as a starting docstring word
+  "D406", # section name ends in new line - is this needed? many docs fail this
+  "D407", # dashed underline missing after section - should I enable this? many docs fail this
```
joaander, line 83: Looking at `setup.cfg` below, it appears that pydocstyle warnings were never enabled. Instead of ignoring all of these `D4*` warnings, consider updating the documentation to follow the numpy style convention that these check.

If there are some exceptional cases that you simply cannot document following the chosen convention, then use `noqa` to ignore those on a case by case basis: https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

joaander (changes requested):
Thanks! I have some suggestions for improvement.

pyproject.toml
```diff
@@ -64,11 +54,36 @@ exclude = '''
...
+    "B",
+    "E",
+    "F",
+    "N",
+    "W", # ruff supports W but, W503 is not supported!!
+    "D", # pydocstyle
+    "I", # isort
+]
+ignore =[
+  "B904", # I think this one doesn't work as intended.
```
janbridley, line 72: Has this been confirmed to work properly?

https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

janbridley (approved):

pyproject.toml
joaander: Indeed, it would be good to raise chained exceptions. Can you give an example of a place where this warning triggers when it should not?

DomFijan: It's complaining about the following two pieces of code, however according to [docs](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) `if` should be inside `except` for this to be triggered:
[Code 1](https://github.com/glotzerlab/coxeter/blob/dcfc9f85d382d3182769de01e86a3a847dfa97f5/coxeter/shapes/base_classes.py#L182C1-L197C38):
```
            try:
                import importlib

                backend = importlib.import_module(f"plato.draw.{backend}")
            except ImportError:
                raise ImportError(
                    f"Backend plato.draw.{backend} could not be imported."
                )
            if scene_kwargs is None:
                scene_kwargs = {}
```
[Code 2](https://github.com/glotzerlab/coxeter/blob/dcfc9f85d382d3182769de01e86a3a847dfa97f5/coxeter/shapes/utils.py#L47C1-L60C55):
```
        try:
            import matplotlib.pyplot as plt
        except ImportError:
            raise ImportError("matplotlib must be installed for plotting.")
        fig = plt.figure()
        if not axes3d:
            ax = fig.subplots()
            ax.set_aspect("equal", "datalim")
        else:
            # This import registers the 3d projection
            from mpl_toolkits.mplot3d import Axes3D  # noqa: F401

            ax = fig.add_subplot(111, projection="3d")
```
Thoughts?

janbridley (approved):

pyproject.toml
joaander: Yes, you should:

```
except ImportError as exception:
    raise ImportError(...) from exception
````
The presence of the `if` in the documentation example is extraneous. Exception chaining is a good practice any time you raise from within an `except` block.

File: pyproject.toml [hunk 3]
```diff
@@ -64,11 +54,36 @@ exclude = '''
...
+  "B028", # This forces usage of stack level in warnings.
```
joaander, line 73: You use `warnings.warn` in coxeter to warn about deprecated properties. I think the `stacklevel=2` option would be helpful for users and the recommendation in the docs is a good one: https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/. This will help users identify which line in *their* code is causing the warning.

joaander (changes requested):
On reviewing the two ignored bugbear checks, I think they are both good recommendations. Please remove the B* ignores and implement the recommended changes to the code.

janbridley (approved):

janbridley:
From `doc/development.rst`:

```
    Alternatively, the tools can be run manually using the commands:

  .. code-block:: bash

      black coxeter/ tests/
      isort coxeter/ tests/
```

Should we update this with commands to manually run ruff?

doc/source/development.rst
```diff
@@ -70,10 +70,12 @@ All code should of course also follow the principles in `PEP 20 <https://www.pyt
     .. code-block:: bash
 
         black coxeter/ tests/
-        isort coxeter/ tests/
+
+        # From the root of the repository
+        ruff check . --fix
```
janbridley, line 75: "Ruff's pre-commit hook should be placed after other formatting tools, such as Black and isort, unless you enable fixes, in which case, Ruff's pre-commit hook should run before Black, isort, and other formatting tools, as Ruff's fix behavior can output code changes that require reformatting."

janbridley: [e40890f](https://github.com/glotzerlab/coxeter/pull/203/commits/e40890f6121d13b331c1bc6924cc33979ec0ce7a) should address this

joaander (approved):
This looks great, thanks!

PR #211, opened 2024-01-18
Improve shape family tests

janbridley:
## Description
More tests were added and current tests were expanded for `test_shape_families.py`

## Motivation and Context
Plane shape families can result in unexpected errors with parameter values close to transition points in shape space. More tests were added to cover this possibility, and integer values were replaced with "exact" floats to verify that transition point shapes can be generated. These tests confirm that the unexpected failures are a result of floating point precision, and not a deeper bug in the implementation. Also note that before #188 these functions return incorrect results rather than raising an error.

## Types of Changes

- [x] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/Credits.rst
+++ b/Credits.rst
@@ -110,6 +110,7 @@ Jen Bradley
 * Fixed error where ``__repr__`` would fail for polyhedra with multiple face types.
 * Increased accuracy of stored data for PlatonicFamily solids
 * Added shape families for Archimedean, Catalan, and Johnson solids.
+* Expanded on tests for shape families
 * Added shape family for prisms and antiprisms.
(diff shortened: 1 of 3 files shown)


DomFijan (commented):
Are 1000 max examples really needed for intermediates? it feels like this can be significantly reduced?

janbridley:
> Are 1000 max examples really needed for intermediates? it feels like this can be significantly reduced?

Hypothesis does sample better with 1000 examples, but the behavior is confirmed with the default 200 samples. I will adjust that down!

DomFijan (approved):

PR #212, opened 2024-01-29
Added new to_hoomd method

janbridley:
## Description

This PR adds a new `to_hoomd()` method that allows for easy use of coxeter objects with Signac. The intended use case for this is saving a shape to a job document, which allows for simple access to properties important to the simulation.

## Motivation and Context

Coxeter works well with simulation tools like `hoomd`, but it would be beneficial to to have a direct export method. While the `gsd_shape_spec` is useful for saving simulation data, it does not contain enough data to initialize a simulation in either MC or MD code.

Additional tests have been added to the relevant test files: `test_sphere.py::test_to_hoomd`, `test_ellipsoid::test_to_hoomd`, `test_polygon::test_to_hoomd`, `test_polyhedron::test_to_hoomd`, and `test_spheropolyhedron::test_to_hoomd`

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


janbridley:
I looped in @cbkerr on this for some feedback on the signac integration.

coxeter/__init__.py
```diff
@@ -21,4 +21,4 @@
 
 __all__ = ["families", "shapes", "from_gsd_type_shapes"]
 
-__version__ = "0.7.0"
+__version__ = "0.8.0"
```
DomFijan, line 24: this should only be changed in PRs where we release a new version?

doc/source/conf.py
```diff
@@ -22,8 +22,8 @@
 author = "Vyas Ramasubramani"
 
 # The full version, including alpha/beta/rc tags
-version = "0.7.0"
-release = "0.7.0"
+version = "0.8.0"
+release = "0.8.0"
```
DomFijan, line 26: version bumps should be only in release PR?

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -267,3 +267,42 @@ def _plato_primitive(self, backend):
             vertices=verts[:, :2],
             radius=self.radius,
         )
+
+    def to_hoomd(self):
+        """Get a json-serializable subset of ConvexSpheropolygon properties.
```
cbkerr, line 272:
```suggestion
        """Get a JSON-serializable subset of ConvexSpheropolygon properties.
```

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -267,3 +267,42 @@ def _plato_primitive(self, backend):
...
+
+        The json-serializable output of the to_hoomd method can be directly imported
+        into data management tools like Signac. This data can then be queried for use in
```
cbkerr, line 275:
```suggestion
        into data management tools like signac. This data can then be queried for use in
```

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -267,3 +267,42 @@ def _plato_primitive(self, backend):
...
+            The rounding radius of the shape.
+        * area (float)
+            The area of the shape.
+        * moment_inertia (list(list))
+            The shape's inertia tensor.
+
+        Returns
+        -------
+        dict
+            Dict containing a subset of shape properties.
```
cbkerr, line 296: This wording raises the question of "why not all properties?" Maybe expand to say "useful shape properties" or something

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -267,3 +267,42 @@ def _plato_primitive(self, backend):
...
+        """
+        old_centroid = self.centroid
+        self.centroid = np.array([0, 0, 0])
```
cbkerr, line 299: I assume this translate the vertices

coxeter/shapes/convex_spheropolygon.py
```diff
@@ -267,3 +267,42 @@ def _plato_primitive(self, backend):
...
+
+        The json-serializable output of the to_hoomd method can be directly imported
+        into data management tools like Signac. This data can then be queried for use in
+        HOOMD simulations. Key naming matches HOOMD integrators: for example, the
+        moment_inertia key links to data from coxeter's inertia_tensor.
+
+        For a ConvexSpheropolygon, the following properties are stored:
+
+        * vertices (list(list)):
+            The vertices of the shape.
```
cbkerr, line 282: Mention that the vertices are also appropriately translated. Could also be mentioned in the main docstring block

cbkerr (commented):
I don't see any problems with the signac aspect, but I didn't run it with signac.

I think the name `to_hoomd` makes it less discoverable. Although it uses the integrator names of HOOMD, it would be much more broadly useful. 

Does this imply that we will add functions like `to_lammps` in the future? I wouldn't support adding extra overhead to support more backends until the need arises.

What if it were called `to_JSON`? People using other tools could change the key names as needed and might not think to search for hoomd.

The only thing making it specific to hoomd is that `inertia_tensor` is mapped to `moment_inertia`. The `to_JSON` function should probably keep coxeter's naming, requring hoomd users to change the key.




**Extra feature idea**: exporting triangulated meshes to rendering tools like fresnel. Not sure what ovito would need. Would this be helpful?

eg I have a function like this where I get the surfaces needed for fresnel.
```
def tris_for_fresnel(vertices):
    poly = coxeter.shapes.ConvexPolyhedron(vertices)
    t=[i for i in poly._surface_triangulation()]
    tris=[]
    for i in t:
        tris.append(i[0])
        tris.append(i[1])
        tris.append(i[2])
    return tris
```

coxeter/shapes/convex_spheropolygon.py
janbridley: This is now mentioned in the main docstring

coxeter/shapes/convex_spheropolygon.py
janbridley: This is correct, and the documentation is updated to reflect that

coxeter/shapes/convex_spheropolygon.py
janbridley: Fixed!

coxeter/shapes/convex_spheropolygon.py
janbridley: Documentation has been updated to be more descriptive

coxeter/shapes/convex_spheropolyhedron.py
```diff
@@ -328,3 +329,38 @@ def _plato_primitive(self, backend):
...
+        dict
+            Dict containing a subset of shape properties required for HOOMD function.
+        """
+        old_centroid = self._polyhedron.centroid
+        self._polyhedron.centroid = np.array([0, 0, 0])
+        data = self.to_json(["vertices", "radius", "volume"])
+        hoomd_dict = _map_dict_keys(data, key_mapping=_hoomd_dict_mapping)
+        hoomd_dict["centroid"] = [0, 0, 0]
+
+        self._polyhedron.centroid = old_centroid
```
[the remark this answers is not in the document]
janbridley: I prefer the current approach because it means to_json always pulls the correct data (and the output does not need to be edited). While there is no functional difference, calling an internal method ensures an extra layer of certainty that the operation is correct

coxeter/shapes/convex_spheropolyhedron.py
```diff
@@ -328,3 +329,38 @@ def _plato_primitive(self, backend):
...
+        moment_inertia key links to data from coxeter's inertia_tensor. Stored values
+        are based on the shape with its centroid at the origin.
+
+        For a ConvexSpheropolyhedron, the following properties are stored:
+
+        * vertices (list(list)):
+            The vertices of the shape.
+        * centroid (list(float))
+            The centroid of the shape.
+            This is set to [0,0,0] to improve HOOMD performance.
```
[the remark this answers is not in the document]
DomFijan: In that case I would slightly reword this to state that hoomd expects 0,0,0 position or something along these lines. Current text implies that this choice is sort-of optional and 0,0,0 is done for performance only? This is not a big issue, though.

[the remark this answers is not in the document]
janbridley: I will update for clarity!

Edit with more details: Convex polyhedra, spheropolyhedra, polygons, and spheropolygon must contain the origin in HOOMD, and performance is maximized when their centroid is at the origin^*. Concave shapes do not necessarily have to contain the origin, but should contain it within their convex hull.

Technically, performance is maximized when the origin centered bounding sphere is of minimal size. We could perform this calculation (and should once #178 is complete), but for now setting the centroid to the origin is equivalent for most high-symmetry shapes.

DomFijan (approved):

janbridley:
Once cbkerr approves these changes I will merge and start working on a new release

cbkerr (approved):
Looks great!

PR #213, opened 2024-01-31
Correct Spheropolyhedron Class

tfteague:
## Description
There is an error in the method via which coxeter currently computes the volumes and surface area of speropolyhedra. Part of the sums of the volume & SA are based on fractions of cylinders that extend from the edges of the base polyhedron. The included angle of this cylinder is equal to the angle between the normals of the two adjoining faces, which is equal to $\pi - \phi$, where $\phi$ is the dihedral (see e.g. equation (9) of Eric, I. M., & Michael, E. (2017). Virial Coefficients and Equations of State for Hard Polyhedron Fluids). Currently the sum is calculated only as $\phi$. While this gives correct results for the test case of a cube ($\pi/2 = \pi - \pi/2$), it does not for non-orthogonal geometries.

In addition, the mean curvature of the spherobody is implemented. The (normalized) integral mean curvature is just the sum of the mean curvature of the base shape and the radius.

## Motivation and Context

- To correct the surface area and volume functions for speropolyhedra
- To add a mean curvature property to the speropolyhedra class

## Types of Changes

- [ ] Documentation update
- [ x] Bug fix
- [ x] New feature

## Checklist:

- [ x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/coxeter/blob/master/CONTRIBUTING.md).
- [ ]x I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/coxeter/blob/master/ContributorAgreement.md).
- [ x] The changes introduced by this pull request are covered by existing or newly introduced tests.

### DIFF
--- a/tests/test_spheropolyhedron.py
+++ b/tests/test_spheropolyhedron.py
@@ -56,6 +56,27 @@ def test_surface_area_polyhedron(convex_cube):
     assert sphero_cube.surface_area == approx(convex_cube.surface_area)
 
 
+@given(radius=floats(0.1, 1))
+def test_mean_curvature(radius):
+    sphero_cube = make_sphero_cube(radius=radius)
+    h_cube = 3 / 4
+    h_sphere = radius
+    assert np.isclose(sphero_cube.mean_curvature, h_cube + h_sphere)
+
+
+def test_mean_curvature_polyhedron(convex_cube, cube_points):
+    """Ensure that zero radius gives the same result as a polyhedron."""
+    sphero_cube = make_sphero_cube(radius=0)
+    assert sphero_cube.mean_curvature == approx(convex_cube.mean_curvature)
+
+
+@given(value=floats(0.1, 1))
+def test_set_mean_curvature(value):
+    sphero_cube = make_sphero_cube(radius=0)
+    sphero_cube.mean_curvature = value
+    assert sphero_cube.mean_curvature == approx(value)
+
+
 @given(r=floats(0, 1.0))
 def test_radius_getter_setter(r):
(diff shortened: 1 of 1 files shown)


coxeter/shapes/convex_spheropolyhedron.py
```diff
@@ -179,6 +183,26 @@ def surface_area(self, value):
...
+        # 1) The mean curvature of the spherical vertex caps, which is just the
+        #    rounding radius of the spheropolyhedron.
+        # 2) The mean curvature of cylindrical wedges along the edges, which sum
+        #    to the mean curvature of the underlying polyhedron
+        h_sphere = self.radius
+        h_cyl = self.polyhedron.mean_curvature
+        return h_cyl + h_sphere
+
+    @mean_curvature.setter
+    def mean_curvature(self, value):
```
janbridley, line 199: Does the change in mean curvature require a change in both the volume of the shape and the rounding radius? And does changing both values at once result in the expected behavior?

tfteague: Not necessarily, no. One could set the rounding radius or change the volume of the underlying polyhedron independently in order to achieve a desired mean curvature. The challenge with that approach would be that not all possible (positive, real) values could be achieved. For example, if one wanted to adjust the curvature by changing the radius alone, then one could not achieve a curvature smaller than that of the underlying polyhedron. Presumably for this reason, the setter methods for volume and surface area for this class scale the object when setting a value. I've tried to match the behavior of the already established functions.

janbridley (approved):

PR #225, opened 2024-06-20
Feature/additional geometry file export methods

josephburkhart:
## Description
I have added a some methods to the `shapes.Polyhedron` class to allow users to export 3D geometry to a handful of common file formats. None of these methods introduce additional dependencies. The following file formats would be supported:
- `.obj` Wavefront OBJ
- `.off` Object File Format
- `.stl` Stereolithography
- `.ply` Polygon File Format
- `.x3d` Extensible 3D
- `.vtk` Legacy VTK
- `.html` HTML (embedded X3D in a simple HTML document)

Note: this is a DRAFT PULL REQUEST. With input from other collaborators, I may still make significant changes to the new methods.

## Motivation and Context
This change could make it significantly easier for users to visualize their geometries and export them for rendering and other publication-related tasks. Currently, users have to hack together their own export functions if they want to export their geometries to, say, blender.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


josephburkhart:
Planning to start drafting some tests tomorrow.

josephburkhart:
In a conversation offline, @janbridley suggested putting this functionality in a separate file, rather than attaching it via methods to the `shapes.Polyhedron` class.

Currently, the implementation only works for polyhedrons, because it requires explicitly defined vertices and faces. If we put export functions into a separate file, it may be reasonable to allow them to accept any subclass of `shapes.Shape3D`. If we wanted to do that, we would need a way to generate vertices and faces to approximate the surfaces of `shapes.Ellipsoid` and `shapes.Sphere`. If any folks have suggestions on previous implementations of such calculations in coxeter's current dependencies, I'd appreciate it. I could implement such calculations myself, but I expect someone else has already done this.

janbridley:
> In a conversation offline, @janbridley suggested putting this functionality in a separate file, rather than attaching it via methods to the `shapes.Polyhedron` class.
> 
> 
> 
> Currently, the implementation only works for polyhedrons, because it requires explicitly defined vertices and faces. If we put export functions into a separate file, it may be reasonable to allow them to accept any subclass of `shapes.Shape3D`. If we wanted to do that, we would need a way to generate vertices and faces to approximate the surfaces of `shapes.Ellipsoid` and `shapes.Sphere`. If any folks have suggestions on previous implementations of such calculations in coxeter's current dependencies, I'd appreciate it. I could implement such calculations myself, but I expect someone else has already done this.

As a simpler alternative, we could have the export code in a separate file that is called as a`Polyhedron` method. This gets us the development and maintenance benefits without requirements additional approximations for rounded shapes. If those approximations are something we wanted, then this approach seems perfect. I do know that mesh approximations of concave curved shapes were brought up offline recently, although generalizations of this sort are nontrivial.

josephburkhart:
pre-commit.ci autofix

coxeter/io.py
```diff
@@ -0,0 +1,280 @@
...
+                shape.centroid[i] -= m
+
+        # Write data
+        vs = shape.vertices
+        file.write(f"solid {shape.__class__.__name__}\n")
+
+        for f in shape.faces:
+            # Decompose face into triangles
+            # ref: https://stackoverflow.com/a/66586936/15426433
+            triangles = [[vs[f[0]], vs[b], vs[c]] for b, c in zip(f[1:], f[2:])]
```
janbridley, line 96: This is a nice little snippet

coxeter/io.py
```diff
@@ -0,0 +1,280 @@
...
+            "xsd:noNamespaceSchemaLocation": "http://www.web3d.org/specifications/x3d-4.0.xsd",
+        },
+    )
+    x3d_scene = ET.SubElement(root, "Scene")
+    x3d_shape = ET.SubElement(
+        x3d_scene, "shape", attrib={"DEF": f"{shape.__class__.__name__}"}
+    )
+
+    x3d_appearance = ET.SubElement(x3d_shape, "Appearance")
+    x3d_material = ET.SubElement(
```
janbridley, line 173:
```suggestion
    _ = ET.SubElement(x3d_shape, "Appearance")
    _ = ET.SubElement(
```

coxeter/io.py
```diff
@@ -0,0 +1,278 @@
...
+    _ = ET.SubElement(x3d_appearance, "Material", attrib={"diffuseColor": "#6495ED"})
+
+    # Geometry data
+    coordinate_indices = list(range(sum([len(f) for f in shape.faces])))
+    prev_index = 0
+    for f in shape.faces:
+        coordinate_indices.insert(len(f) + prev_index, -1)
+        prev_index += len(f) + 1
+
+    coordinate_points = [v for f in shape.faces for i in f for v in shape.vertices[i]]
```
janbridley, line 182: Consider rewriting this with numpy functions - ravel, etc. Should be possible to simplify!

josephburkhart:
Quick question for @janbridley and/or @DomFijan: in the "Raises" section of a docstring, should I specify only errors that could directly be raised by the function, or should I also include errors that could be raised by other functions that that function calls? 

For example, in the new [`Polyhedron.save()`](https://github.com/glotzerlab/coxeter/blob/927b2cd467a19a8fc67cb8c170ac4d1a2fc44c9b/coxeter/shapes/polyhedron.py#L1025-L1059), I indicate that it raises `ValueError`. Should I also specify that it raises `OSError`, since that error can be raised by the functions that it calls under the hood (`io.to_obj()`, `io.to_off()`, etc)?

janbridley:
> Quick question for @janbridley and/or @DomFijan: in the "Raises" section of a docstring, should I specify only errors that could directly be raised by the function, or should I also include errors that could be raised by other functions that that function calls? 
> 
> 
> 
> For example, in the new [`Polyhedron.save()`](https://github.com/glotzerlab/coxeter/blob/927b2cd467a19a8fc67cb8c170ac4d1a2fc44c9b/coxeter/shapes/polyhedron.py#L1025-L1059), I indicate that it raises `ValueError`. Should I also specify that it raises `OSError`, since that error can be raised by the functions that it calls under the hood (`io.to_obj()`, `io.to_off()`, etc)?

I think because the called methods are documented to raise an OSerror,we should probably mention that here too! Just copy the text from one of the methods

coxeter/io.py
```diff
@@ -0,0 +1,293 @@
+# Copyright (c) 2021 The Regents of the University of Michigan
```
joaander, line 1:
```suggestion
# Copyright (c) 2024 The Regents of the University of Michigan
```
Consider using https://github.com/glotzerlab/fix-license-header to keep all the license headers up to date.

janbridley:
Fixed the circular import - not we just needs tests and we are good to merge!

janbridley:
Merged in the new CI to verify it's working well. svg3d needs some work and I don't want to block this PR, so I will add that in later

tests/test_io.py
```diff
@@ -0,0 +1,103 @@
...
+
+from coxeter import io
+from coxeter.shapes import ConvexPolyhedron, Polyhedron
+
+
+def compare_text_files(file_path_1, file_path_2):
+    """Check whether two text files have identical contents, ignoring different
+    newline characters."""
+    with open(file_path_1) as file1, open(file_path_2) as file2:
+        file1_contents, file2_contents = file1.read(), file2.read()
```
janbridley, line 22:
```suggestion
        file1_contents, file2_contents = file1.read(), file2.read()
        for i, char in file1_contents:
            print(char, file2_contents[i])
```

josephburkhart:
@janbridley all the tests are now passing, thanks to @joaander 's help. As the code currently stands, we would need to re-generate the control files with every version release, since the export functions include the coxeter version number in some of the files. There's an `if __name__ == "__main__"` block in `tests/test_io.py` that generates new control files, so a call to this file would need to be added to the version release workflow. If that sounds like too much of a hassle, I could add in some functionality that ignores coxeter version numbers during the file comparison check, but this might complicate adding support for new file formats in the future.

janbridley (approved):
Looks good! Thank you for this

janbridley:
> @janbridley all the tests are now passing, thanks to @joaander's help. As the code currently stands, we would need to re-generate the control files with every version release, since the export functions include the coxeter version number in some of the files. There's an `if __name__ == "__main__"` block in `tests/test_io.py` that generates new control files, so a call to this file would need to be added to the version release workflow. If that sounds like too much of a hassle, I could add in some functionality that ignores coxeter version numbers during the file comparison check, but this might complicate adding support for new file formats in the future.

It should be fairly simple to automate running that file with the CI. I'll add this in the next release PR. I think the comparison function is good for now, and we can always replace the contents of that function in the future for more detailed/version-agnostic feedback.

PR #231, opened 2024-07-08
Use uv in CI

janbridley:
## Description

Testing for coxeter is very extensive, so the CI is slow. Updating the CI framework as well as the number of pytest threads should aid in this. I also added a (manually triggered) action to update pipfiles, and updated the current pipfiles to work with `pip sync`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:


.github/workflows/CI.yaml
```diff
@@ -29,9 +29,10 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - uses: glotzerlab/workflows/setup-uv
```
janbridley, line 32:
```suggestion
    - uses: glotzerlab/workflows/setup-uv@ea2e25d07af862a1c696a932c2bd6b242d142049
```

.github/workflows/update_pipfiles.yaml
```diff
@@ -0,0 +1,29 @@
...
+    strategy:
+      matrix:
+        python-version: ["3.8","3.9","3.10","3.11","3.12"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - uses: glotzerlab/workflows/setup-uv
```
janbridley, line 24:
```suggestion
    - uses: glotzerlab/workflows/setup-uv@ea2e25d07af862a1c696a932c2bd6b242d142049
```

.github/workflows/CI.yaml
```diff
@@ -29,9 +29,10 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - uses: glotzerlab/workflows/setup-uv@ea2e25d07af862a1c696a932c2bd6b242d142049
```
joaander, line 32:
```suggestion
    - uses: glotzerlab/workflows/setup-uv@ea2e25d07af862a1c696a932c2bd6b242d142049 # 0.2.0
      with:
        lockfile: ".github/requirements${{matrix.python-version}}.txt"
```
And delete the now duplicate "Set up Python environment" step.

.github/workflows/CI.yaml
```diff
@@ -40,4 +41,4 @@ jobs:
         python -m pip install . --progress-bar off
```
joaander, line 41:
```suggestion
        uv pip install . --no-build-isolation --no-deps -v
```
Install with the dependencies explicitly provided by the synced environment.

.github/workflows/CI.yaml
```diff
@@ -40,4 +41,4 @@ jobs:
...
     - name: Test with pytest
       run: |
-        python -m pytest -v
+        python -m pytest -v -n auto
```
joaander, line 44: Nice! So `pytest-xdist` was previously installed but not actually being used?

joaander (changes requested):
Thanks! I have suggestions to improve this.

.github/workflows/CI.yaml
janbridley: Looks like I was missing it actually! Somewhere along the way I forgot to commit the file.

.github/workflows/CI.yaml
janbridley: This doesn't work in its current implementation because miniball doesn't have binaries. See [glotzerlab/workflows#8](https://github.com/glotzerlab/workflows/issues/8)

.github/workflows/CI.yaml
```diff
@@ -29,15 +29,16 @@ jobs:
...
       run: |
-        python -m pip install -r .github/requirements${{matrix.python-version}}.txt
+        uv pip sync ".github/requirements${{matrix.python-version}}.txt" --reinstall --system
     - name: Install package
       run: |
         python --version
         python -c "import numpy; print('numpy', numpy.__version__)"
         python -c "import scipy; print('scipy', scipy.__version__)"
-        python -m pip install . --progress-bar off
+        uv pip install . --no-build-isolation --no-deps -v
```
janbridley, line 41:
```suggestion
        uv pip install . --no-build-isolation --no-deps --system -v
```

.github/workflows/CI.yaml
janbridley:
```suggestion
    - uses: glotzerlab/workflows/setup-uv@ea2e25d07af862a1c696a932c2bd6b242d142049 # 0.2.0
      with:
      	onlybinary: "false"
        lockfile: ".github/requirements${{matrix.python-version}}.txt"
```

.github/workflows/update_pipfiles.yaml
joaander, line 1: This looks right now. The "Run workflow" button does not appear in https://github.com/glotzerlab/coxeter/actions/workflows/update_pipfiles.yaml so I'm not sure how to test this.

joaander (approved):
This looks good now.

PR #236, opened 2024-08-26
Untrack ipynb in linguist

janbridley:
## Motivation and Context
Github linguist autoselects the language for a repository. For coxeter, this results in the following:

image

Listing the package as primarily jupyter notebook could be confusing,and as far as I know there is no GLSL code. Updating the .gitattributes is a minor fix but a useful one nonetheless

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:


janbridley:
> It appears that https://github.com/glotzerlab/coxeter/blob/master/paper/figure1/shape_images/tri.txt.vert is the GLSL file:
> 
> https://github.com/github-linguist/linguist/blob/d190ad88af555bb2ee213aa64205f2b9a97a077d/lib/linguist/languages.yml#L2230-L2256

It should be possible to tell linguist to only run on the source directory - I'll try that

.gitattributes
```diff
@@ -0,0 +1 @@
+coxeter/* linguist-vendored
```
janbridley, line 1:
```suggestion
paper/** linguist-vendored
```

janbridley:
```bash
$ github-linguist
100.00% 360976     python
```

linguist-vendored untracks a folder in linguist, and subfolders must be added with parent/** instead of parent/*. This is all set now!

joaander (approved):
Thanks!

janbridley:
Merge appears to be locked to admins only - I could override, but is there a systematic setting to change to fix this?

PR #310, opened 2025-12-18
Distance

janbridley:
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/coxeter/blob/main/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/coxeter/blob/main/ContributorAgreement.md).
- [ ] I have updated the [changelog](https://github.com/glotzerlab/coxeter/blob/main/ChangeLog.rst) and the [credits](https://github.com/glotzerlab/coxeter/blob/main/Credits.rst).


coxeter/shapes/_distance2d.py
```diff
@@ -0,0 +1,832 @@
...
+    -------
+        np.ndarray: displacements (n_points, 3)
+    """
+    vert_point_vect = vert - point
+    face_unit = face_normal / np.linalg.norm(
+        face_normal
+    )  # unit vector of the normal of the polygon
+    disp = (
+        np.expand_dims(np.sum(vert_point_vect * face_unit, axis=1), axis=1) * face_unit
+    )  # *(-1)
```
janbridley:
```suggestion
    )
```

coxeter/shapes/_distance2d.py
rynoliphant, line 1: Yes, I think this is right (and better too, there could have been issues with the old code if someone initialized a polygon with the vertices out of order). Plus, tests are still passing

tests/test_spheropolyhedron.py
```diff
@@ -158,3 +159,132 @@ def test_to_hoomd(poly, r):
...
+    true_displacements = np.array(
+        [[0, 0, -1], [-1, -1, 1], [-1, 0, 0], [0, -1, -1], [0, 0, -1], [0, 0, -2]]
+    )
+    true_displacements = true_displacements - radius * (
+        true_displacements
+        / np.expand_dims(np.linalg.norm(true_displacements, axis=1), axis=1)
+    )
+
+    np.testing.assert_allclose(distances, true_distances)
+    np.testing.assert_allclose(displacements[1:7], true_displacements)
```
janbridley: @rynoliphant do you have the other displacement values written out?

rynoliphant: The reason the other displacement values weren't written out was because there are multiple possible displacements to the surface (besides point [4,3,3], which I think actually has only one valid displacement). Point [3,3,3] is in the center of the spheropolyhedron and has 6 valid displacements (one to each face), point [4,4,4] sits on a vertex of the inner cube and has infinite valid displacements, point [4,4,3] sits on an edge of the inner cube and has infinite valid displacements, and point [4,3,3] sits on a face of the inner cube and has 1 valid displacement (which I can add, it should be [radius, 0,0] )

janbridley: Noted -- what does the internal code do in this case? Is it left up to whichever is closer in floating point?

janbridley:
Looks great Ryn! Do you want to add yourself to the credits? If not I can write in a blurb

rynoliphant:
If writing a blurb doesn't take much of your time, I would greatly appreciate it! If not though, I can write something

